PR #7, opened 2021-01-06
fix broken CryptoKey handling on web if running in release mode (dart2js)

tauu:
This PR fixes an error in the CryptoKey handling, that occurs only if the package is compiled using the dart2js compiler which is used by flutter in release and profile mode.

The error does not show up in any of the unit tests, presumably because the tests for the web platform are run using the dart devc compiler, which performs differently in some cases. To illustrate the issue, an example flutter program is provided, which just imports a raw base64 encoded AES GCM key.
https://github.com/tauu/webcrypto_dart2js_error
If the app is run in debug mode, using
```bash
flutter run -d chrome
```
and the button next to the text field is clicked, importing the AES GCM key works fine. But if it is run in profile mode (or release mode)
```bash
flutter run -d chrome --profile
``` 
an exception occurs:
```
import failed: NoSuchMethodError: method not found: 'get$type' (J.getInterceptor$x(...).get$type is not a function) 

 TypeError: J.getInterceptor$x(...).get$type is not a function
    at Object.get$type$x (http://localhost:54102/main.dart.js:8393:43)
...
```
The origin of the exception is https://github.com/google/webcrypto.dart/blob/e5c87ad597e21226180f2f7d0391f3c1104af87d/lib/src/impl_js/impl_js.utils.dart#L175 where accessing the type attribute of the imported CryptoKey fails. Adding a line verifying the type of `k` similar to
```
if (k is subtle.CryptoKey) {
  print('k is CrytpoKey')
}
```
will reveal that k is not considered to be of type CrytoKey.

This is a known problem with dart2js if a call returns a promise, which yields a native type,  see e.g. https://github.com/dart-lang/sdk/issues/42796 . Usually, this is resolved by annotating the call creating the promise or the returned dart class, but these annotations are only available in the sdk itself. Consequently fixing the custom CryptoKey class is not possible at the moment (at least to my knowledge).

The issue is fixed by dropping the custom CryptoKey class in favor of the CryptoKey class defined by dart:html: https://api.dart.dev/stable/2.10.4/dart-html/CryptoKey-class.html . This  class is properly annotated to circumvent the issue above. The branch https://github.com/tauu/webcrypto_dart2js_error/tree/dart2js-cryptokey-fix of the aforementioned example app overrides the webcrypto dependency with the version provided with this PR and works also in release and profile mode.

### DIFF
--- a/example/test_driver/app_test.dart
+++ b/example/test_driver/app_test.dart
@@ -33,7 +33,8 @@ void main() {
       await driver.tap(importKeyButtonFinder);
 
       // The outout should contain the importet key now.
-      expect(await driver.getText(keyOutputFinder), '{"kty":"oct","alg":"A128GCM","k":"3nle6RpFx77jwrksoNUb1Q"}');
+      expect(await driver.getText(keyOutputFinder),
+          matches('.*"3nle6RpFx77jwrksoNUb1Q".*'));
     });
 
     test('no aes key generated', () async {
@@ -46,7 +47,8 @@ void main() {
       await driver.tap(generateKeyButtonFinder);
 
       // The outout should contain the importet key now.
-      expect(await driver.getText(genKeyOutputFinder), matches('{"kty":"oct","alg":"A256CBC","k":"[^"]{43}"}'));
+      expect(await driver.getText(genKeyOutputFinder),
+          matches('.*"alg":"A256CBC".*'));
     });
   });
 }
--- a/example/test_driver/app_test.dart
+++ b/example/test_driver/app_test.dart
@@ -1,0 +1,52 @@
+// Imports the Flutter Driver API.
+import 'package:flutter_driver/flutter_driver.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('Webcrypto Example App', () {
+    final keyOutputFinder = find.byValueKey('KeyOutput');
+    final importKeyButtonFinder = find.byValueKey('ImportRawKey');
+    final genKeyOutputFinder = find.byValueKey('GenKeyOutput');
+    final generateKeyButtonFinder = find.byValueKey('GenerateKey');
+
+    FlutterDriver driver;
+
+    // Connect to the Flutter driver before running any tests.
+    setUpAll(() async {
+      driver = await FlutterDriver.connect();
+    });
+
+    // Close the connection to the driver after the tests have completed.
+    tearDownAll(() async {
+      if (driver != null) {
+        driver.close();
+      }
+    });
+
+    test('no imported key', () async {
+      // Verify that no key has been imported in the begining.
+      expect(await driver.getText(keyOutputFinder), "-");
+    });
+
+    test('import raw aes key', () async {
+      // Tap the button to import the key.
+      await driver.tap(importKeyButtonFinder);
+
+      // The outout should contain the importet key now.
+      expect(await driver.getText(keyOutputFinder), '{"kty":"oct","alg":"A128GCM","k":"3nle6RpFx77jwrksoNUb1Q"}');
+    });
+
+    test('no aes key generated', () async {
+      // Verify that no key has been generated in the begining.
+      expect(await driver.getText(genKeyOutputFinder), "-");
+    });
+
+    test('generate aes cbc key', () async {
+      // Tap the button to import the key.
+      await driver.tap(generateKeyButtonFinder);
+
+      // The outout should contain the importet key now.
+      expect(await driver.getText(genKeyOutputFinder), matches('{"kty":"oct","alg":"A256CBC","k":"[^"]{43}"}'));
+    });
+  });
+}
--- a/lib/src/crypto_subtle.dart
+++ b/lib/src/crypto_subtle.dart
@@ -25,7 +25,7 @@ import 'dart:convert' show jsonDecode;
 import 'dart:html' show DomException, CryptoKey;
 import 'jsonwebkey.dart' show JsonWebKey;
 export 'jsonwebkey.dart' show JsonWebKey;
-export 'dart:html' show DomException;
+export 'dart:html' show DomException, CryptoKey;
 
 /// Minimal interface for promises as returned from the browsers WebCrypto API.
 @JS('Promise')
--- a/lib/src/crypto_subtle.dart
+++ b/lib/src/crypto_subtle.dart
@@ -22,7 +22,7 @@ import 'package:js/js.dart';
 import 'package:js/js_util.dart' as js_util;
 import 'dart:typed_data';
 import 'dart:convert' show jsonDecode;
-import 'dart:html' show DomException;
+import 'dart:html' show DomException, CryptoKey;
 import 'jsonwebkey.dart' show JsonWebKey;
 export 'jsonwebkey.dart' show JsonWebKey;
 export 'dart:html' show DomException;
@@ -68,30 +68,6 @@ Uint8List bigIntToUint8ListBigInteger(BigInt integer) {
   // supports 3 and 65537, so supporting other numbers is a low priority.
   // https://chromium.googlesource.com/chromium/src/+/43d62c50b705f88c67b14539e91fd8fd017f70c4/components/webcrypto/algorithms/rsa.cc#286
   throw UnimplementedError('Only supports 65537 and 3 for now');
-}
-
-/// Minimal interface for the CryptoKey type.
-@JS('CryptoKey')
-class CryptoKey {
-  /// Returns the _type_ of this key, as one of:
-  ///  * `'private'`
-  ///  * `'public'`
-  ///  * `'secret'`
-  external String get type;
-
-  /// True, if this key can be extracted.
-  external bool get extractable;
-
-  /// Ways in which this key can be used, list of one or more of:
-  ///  * `'encrypt'`,
-  ///  * `'decrypt'`,
-  ///  * `'sign'`,
-  ///  * `'verify'`,
-  ///  * `'deriveKey'`,
-  ///  * `'deriveBits'`,
-  ///  * `'wrapKey'`,
-  ///  * `'unwrapKey'`.
-  external List<String> get usages;
 }
 
 /// Interface for the [CryptoKeyPair][1].
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -38,6 +38,8 @@ dev_dependencies:
   pedantic: ^1.9.0
   flutter_test:
     sdk: flutter
+  flutter_driver:
+    sdk: flutter
 
 flutter:
   plugin:
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -38,8 +38,6 @@ dev_dependencies:
   pedantic: ^1.9.0
   flutter_test:
     sdk: flutter
-  flutter_driver:
-    sdk: flutter
 
 flutter:
   plugin:
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -14,7 +14,7 @@
 
 import 'package:flutter/material.dart';
 import 'dart:async';
-import 'dart:convert' show utf8;
+import 'dart:convert' show utf8, base64Decode, jsonEncode;
 
 import 'package:convert/convert.dart' show hex;
 import 'package:webcrypto/webcrypto.dart';
@@ -29,15 +29,18 @@ class MyApp extends StatefulWidget {
 }
 
 class _MyAppState extends State<MyApp> {
-  String _output = '-';
-  final _input = TextEditingController();
+  String _hash = '-';
+  String _importedKey = '-';
+  String _generatedKey = '-';
+  final _inputPlain = TextEditingController();
+  final _inputRawKey = TextEditingController(text: "3nle6RpFx77jwrksoNUb1Q==");
 
   @override
   Widget build(BuildContext context) {
     return MaterialApp(
       home: Scaffold(
         appBar: AppBar(
-          title: const Text('SHA-1 Hashing'),
+          title: const Text('Webcrypto'),
         ),
         body: Container(
           padding: EdgeInsets.all(10),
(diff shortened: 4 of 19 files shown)


jonasfj:
I see no significant downside to using `CryptoKey` from `dart:html`.

It's been quite a while since I wrote these JS binding parts, and I recall I did some hacks to workaround various issues I didn't fully investigate at the time. In any case, I'd be happy to get rid of hacks like duplicating classes from `dart:html`.

I have a setup to run tests on various platforms, but at the moment it's all manual, so I'll test this out manually before merging.

tauu:
No problem, thanks for the very fast review! 👍

In case you want to remove more duplicated code, another opportunity turned up during the research: The `promiseAsFuture` function https://github.com/google/webcrypto.dart/blob/e5c87ad597e21226180f2f7d0391f3c1104af87d/lib/src/crypto_subtle.dart#L42 can be replaced by `promiseToFuture` from dart:js_util 
https://api.dart.dev/stable/2.10.4/dart-js_util/promiseToFuture.html

jonasfj:
@tauu, hmm, as I recall it I had to make my own `promiseToFuture` function because the one from `dart:js_util` didn't move the callback to the current zone.. Maybe there is another way to wrap it.
But as I recall it there was problems with exceptions, but this might be worth exploring.

In any case, I think we need to dig deeper here, I have problems running:
```
flutter pub run test -p chrome test/webcrypto/hmac_test.dart
```
(I suspect the promise isn't getting resolved somehow, we might need to make a smaller test case)

I guess this test isn't critical for Flutter Web support, but it would be nice if it also works here. As we're using the same mechanism to test on Firefox and Safari.

tauu:
Sorry for missing that during my checks! I can reproduce the errors and am still investigating the issue.

tauu:
The problem seems to reside with the generateKey function, which can return either a CryptoKey or a CryptoKeyPair. Maybe the compiler has some problems handling the return value.

jonasfj:
> The problem seems to reside with the generateKey function, which can return either a CryptoKey or a CryptoKeyPair. Maybe the compiler has some problems handling the return value.

hmm, interesting, because we have two declarations of the method for exactly this purpose.

jonasfj:
@tauu, can we perhaps make a minimal example that only relies on `crypto_subtle.dart` to produce the error, and ideally some simple commands to reproduce it?

tauu:
@jonasfj that should be possible, I'll give it a try :)

tauu:
@jonasfj here you go:
```dart
import 'dart:html';

import 'package:test/test.dart';
import 'package:webcrypto/src/crypto_subtle.dart';

void main() {
  test('generate key', () async {
    final k = await promiseAsFuture(generateKey(
      Algorithm(
        name: 'AES-CBC',
        length: 256,
      ),
      true,
      ['encrypt', 'decrypt'],
    ));
    expect(k is CryptoKey, true);
  });
}
```
This test will timeout after 30 seconds just like the other already defined tests.
Some further experiments revealed that changing the return type of the generateKey function from CryptoKey to dynamic:
```dart
@JS('crypto.subtle.generateKey')
external Promise<dynamic> generateKey(
  Algorithm algorithm,
  bool extractable,
  List<String> usages,
);
```
will fix the timeout problem, but then k is CryptoKey yields false. Casting the return value to CryptoKey with a slightly modified test
```dart
import 'dart:html';

import 'package:test/test.dart';
import 'package:webcrypto/src/crypto_subtle.dart';

void main() {
  test('generate key', () async {
    final k = await promiseAsFuture(generateKey(
      Algorithm(
        name: 'AES-CBC',
        length: 256,
      ),
      true,
      ['encrypt', 'decrypt'],
    ));
    try {
      final kc = k as CryptoKey;
    } catch (e) {
      print(e);
    }
    expect(k is CryptoKey, true);
  });
}
```
will produce the following exception:
```
TypeError: Instance of 'CryptoKey': type 'UnknownJavaScriptObject' is not a subtype of type 'CryptoKey'
```
Sooo ... the returned type is of the type CryptoKey but at the same time it is not? Did CryptoKey just turn into Schroedinger's type?
Let us dig furhter. Let's check if we can read the type property of `k` and see if we actually got a correct return value.
```dart
import 'dart:html';

import 'package:js/js_util.dart';
import 'package:test/test.dart';
import 'package:webcrypto/src/crypto_subtle.dart';

void main() {
  test('generate key', () async {
    final k = await promiseAsFuture(generateKey(
      Algorithm(
        name: 'AES-CBC',
        length: 256,
      ),
      true,
      ['encrypt', 'decrypt'],
    ));
    print(getProperty(k,"type"));
    try {
      final kc = k as CryptoKey;
    } catch (e) {
      print(e);
    }
    expect(k is CryptoKey, true);
  });
}
```
This will print secret AND to my astonishment no exception. In fact the test will pass now and k is of type CryptoKey. The return type of generateKey can also be changed from dynamic to CryptoKey and the test still passes.
All tests will pass if the following single line is added to the functions _generateKey and _importKey in impl_js_utils.dart :
```dart
final type = getProperty(k,"type");
```
Obviously this is not exactly an ideal way to fix the issue, as it might break in the future if  compiler optimizations change. Nevertheless if you consider it to be ok as a temporary fix, I will be happy to add the changes to the PR.

jonasfj:
Hmm, reading a bit about [package:js](https://pub.dev/packages/js/versions/0.6.3-nullsafety.3) it says:
> This library supersedes dart:js, so don't import dart:js.

Hence, I wonder if it's wise to use the types from `dart:html` at all, maybe we should get rid of `DomException` too.

tauu:
Interesting, so looking at the soruce of package:js https://github.com/dart-lang/sdk/blob/master/pkg/js/lib/js.dart , I get the feeling, that for the most part, it just hides some of the functionality of dart:js .

jonasfj:
I spend a bit of time playing with `flutter drive` as it seems like that'll allow me to run tests in release mode on web.

https://flutter.dev/docs/cookbook/testing/integration/introduction#6b-web

But maybe we should just start by patching the `example/`, such that we can reproduce the original issue by running it in release mode.

tauu:
Would the example app showing the initial error be ok for that? https://github.com/tauu/webcrypto_dart2js_error
In that case I could add the code to the existing example. :-)

tauu:
The example app has been update to include a unit test, which triggers the original issue.

Something else that turned up due to the test: The issues reported e.g. by
```
flutter pub run test -p chrome test/webcrypto/hmac_test.dart
```
or the simple example test mentioned in the previous post do not seem to be triggered in release mode.

tauu:
After thinking about the issue a bit more, the cleanest way to fix it, might be to get all the `crypto.subtle.*` functions into `dart:html`. Functions such as `generateKey`, which return one of multiple types, are annotated there to reflect this behavior. Here is a small example for one of the native apis from `dart:html` :
```dart
@Native("ExtendableMessageEvent")
class ExtendableMessageEvent extends ExtendableEvent {
  @Creates('Client|ServiceWorker|MessagePort')
  @Returns('Client|ServiceWorker|MessagePort|Null')
  Object? get source native;
}
```
What do you think about that?

jonasfj:
> What do you think about that?

Moving it into `dart:html` would be nice, but I'm not sure that's something we can just do.
I borderline suspect it's generated from IDL or something, it's certainly a much harder target to move.

This seems like a bug in the compiler, we should make sure it's reported, and find a way to work around... Using `getProperty` is one way..
Maybe we can just use two annocation `@JS('crypto.subtle.verify')` and `@JS('window.crypto.subtle.verify')` to make the compiler think they are indeed different methods :see_no_evil: 

If the problem really is that the compiler/optimizer somehow merges the two functions but breaks types when doing so, that could be a work around.

tauu:
> Moving it into dart:html would be nice, but I'm not sure that's something we can just do.
> I borderline suspect it's generated from IDL or something, it's certainly a much harder target to move.
Yes, it is generated from IDL. After some investigation into `dart:html`, the IDL used for generation contains all the crypto.suble api function, but the output of the generator for this api is suppressed. A reason for the suppression was not found so far. Removing the suppression generates code that looks pretty good, but some api functions are still missing. No idea why that is the case. :-/

Agreed, a bug report and a workaround will be easier. :-)

pubspec.yaml
```diff
@@ -38,6 +38,8 @@ dev_dependencies:
   pedantic: ^1.9.0
   flutter_test:
     sdk: flutter
+  flutter_driver:
+    sdk: flutter
```
jonasfj, line 42: Why this?

lib/src/crypto_subtle.dart
```diff
@@ -22,7 +22,7 @@ import 'package:js/js.dart';
 import 'package:js/js_util.dart' as js_util;
 import 'dart:typed_data';
 import 'dart:convert' show jsonDecode;
-import 'dart:html' show DomException;
+import 'dart:html' show DomException, CryptoKey;
 import 'jsonwebkey.dart' show JsonWebKey;
 export 'jsonwebkey.dart' show JsonWebKey;
 export 'dart:html' show DomException;
```
jonasfj, line 28: Let's add `CryptoKey` here, so we don't have to update `subtle.CryptoKey` everywhere else..

jonasfj:
Okay, I managed to reproduce the original issue using the example app changes here and:
`flutter drive -d web-server --release --target test_driver/app.dart`
(assuming I have chromedriver running on port 4444).

example/test_driver/app_test.dart
```diff
@@ -0,0 +1,52 @@
...
+      // Verify that no key has been imported in the begining.
+      expect(await driver.getText(keyOutputFinder), "-");
+    });
+
+    test('import raw aes key', () async {
+      // Tap the button to import the key.
+      await driver.tap(importKeyButtonFinder);
+
+      // The outout should contain the importet key now.
+      expect(await driver.getText(keyOutputFinder), '{"kty":"oct","alg":"A128GCM","k":"3nle6RpFx77jwrksoNUb1Q"}');
```
jonasfj, line 36: This fails for me, maybe just:
```suggestion
      expect(await driver.getText(keyOutputFinder),
          contains('3nle6RpFx77jwrksoNUb1Q'));
```

Some browsers might include other values and change ordering of JSON.
We don't need an accurate test here, just hint that it isn't totally wrong :D

example/test_driver/app_test.dart
```diff
@@ -0,0 +1,52 @@
...
+      // Verify that no key has been generated in the begining.
+      expect(await driver.getText(genKeyOutputFinder), "-");
+    });
+
+    test('generate aes cbc key', () async {
+      // Tap the button to import the key.
+      await driver.tap(generateKeyButtonFinder);
+
+      // The outout should contain the importet key now.
+      expect(await driver.getText(genKeyOutputFinder), matches('{"kty":"oct","alg":"A256CBC","k":"[^"]{43}"}'));
```
jonasfj, line 49: Again, it might be more robust to test:
```suggestion
expect(await driver.getText(genKeyOutputFinder),
          contains('A256CBC'));
```
(we could also parse JSON and do more accurate tests, but it's probably not necessary)

Side note: max linelength is 80 in this repo, everything is dartfmt formatted.

jonasfj:
Note. The tests are still broken when using `flutter pub run test -p chrome`

jonasfj:
Played around with this bit, but I can't reproduce the original issue using `dart2js -O4 runalltest.dart -o output.js`,then loading `output.js` into a HTML page on localhost.

With `runalltest.dart` as:
```dart
import 'testrunners.dart';
import 'package:test/test.dart';

void main() => runAllTests(test: test);
```

So something in Flutter Web release / profile mode is breaking this.

pubspec.yaml
tauu: Sorry, I must have added this also to the package pubspec.yml by accident. The dependency has been removed. :-)

lib/src/crypto_subtle.dart
tauu: Good idea! Code has been changed accordingly. :)

example/test_driver/app_test.dart
tauu: You are totally right, there is no need for an accurate test in this case. The test has been adjusted.

example/test_driver/app_test.dart
tauu: Same as before, the test has been adjusted. :-)

dartfmt has also been applied.

tauu:
The PR now also contains the suggested workaround using `getProperty(k,"type")` workaround suggested above. I have not added it earlier as we were still discussing the issue. Could you check again if 
- flutter pub run test -p chrome
- flutter run -d web-server --release
are still broken? On my machine both work using the workaround. 

What is currently not working for some reason is:
- flutter drive --target=test_driver/app.dart --browser-name=chrome
(without --release, with --release it is fine) Surprisingly the app works fine though if it is run in debug mode using `flutter run -d chrome`. :-/

jonasfj:
I filed https://github.com/dart-lang/sdk/issues/44664

I'll take a look at this, but I suspect we'll have accept weird bugs until the compiler bugs are fixed.

JohnGalt1717:
I think that this issue is related:

With 0.5.2 with: AesGcmSecretKey.generateKey(256) I get Expected a value of type 'JSObject<CryptoKey>', but got one of type 'CryptoKey'


Is there a fix or workaround?

tauu:
You can use this pull request as a workaround by setting the following dependency:
```yaml
dependencies:
  webcrypto:
    git:
      url: https://github.com/tauu/webcrypto.dart
      ref: 0.5.2-fix
 ```
 
 In its current state it is not a perfect fix, but it should work.

jhancock4d:
@tauu for what   it's worth I was also able to fix my issue by removing the CryptoKey from subtle.dart and using the html.CryptoKey implementation.

I'm using those files directly, because with the latest version of flutter, the android implementation here freaks out with gradle errors even when not targeting android for debug/deploy.

It might be nice to get a clean web implementation and have that as a separate dart package that this consumes for web too. (and encrypt/decrypt example would be nice too)

JohnGalt1717:
@tauu for what it's worth I was also able to fix my issue by removing the CryptoKey from subtle.dart and using the html.CryptoKey implementation.
I'm using those files directly, because with the latest version of flutter, the android implementation here freaks out with gradle errors even when not targeting android for debug/deploy.
It might be nice to get a clean web implementation and have that as a separate dart package that this consumes for web too. (and encrypt/decrypt example would be nice too)

jonasfj:
I think this is fixed with https://github.com/google/webcrypto.dart/pull/26

PR #17, opened 2021-03-10
Use package:ffigen

dcharkes:
This PR replaces the hand-written bindings with ones generated with `package:ffigen`.

Some details:

* `Byte` and `Data` changed to `Uint8` and `Void` respectively, removing some casts.
* Bindings for BoringSSL and WebCryptoDartDL are generated separately.
* Bindings for BoringSSL are in `lib/src/third_party/boringssl` accompanied by a LICENSE file.
* Code is kept as much in the same place as possible to keep the diff small. (Maybe we should consider moving some stuff around.)


lib/src/boringssl/bindings/ffigen.yaml
```diff
@@ -0,0 +1,53 @@
+name: WebCryptoDartDL
+description: 'Bindings to src/webcrypto_dart_dl.h.'
+output: 'lib/src/boringssl/bindings/generated_bindings.dart'
+headers:
+  entry-points:
+    - src/webcrypto_dart_dl.h
+    - src/webcrypto.c
+compiler-opts: '-Ithird_party/dart-sdk/src/runtime/ -Ithird_party/boringssl/src/include -I/Library/Developer/CommandLineTools/usr/lib/clang/12.0.0/include'
```
jonasfj, line 8: Can `-I/Library/Developer/CommandLineTools/usr/lib/clang/12.0.0/include` be put in a platform specific option somewhere?

third_party.LICENSE
```diff
@@ -0,0 +1,5 @@
+For license on third_party/ contents, see:
```
jonasfj, line 1: Let's drop this file, seems we won't need it :D

tool/generate_symbols_table.dart
```diff
@@ -103,6 +103,13 @@ Future<void> main() async {
     '  }',
     '}',
     '',
+    'extension SymName2 on String {',
```
jonasfj, line 106: Might it not be better to do a static method `SymFromString(String) -> Sym`?

I wish we could static extensions like `Sym.fromString`, hehe..

But an extension on `String` seems a bit dangerous.

lib/src/third_party/boringssl/ffigen.yaml
```diff
@@ -0,0 +1,264 @@
...
+    evp_pkey_asn1_method_st: EVP_PKEY_ASN1_METHOD
+    evp_pkey_ctx_st: EVP_PKEY_CTX
+    evp_pkey_st: EVP_PKEY
+    hmac_ctx_st: HMAC_CTX
+    rsa_meth_st: RSA_METHOD
+    rsa_st: RSA
+  dependency-only: opaque
+functions:
+  include:
+  - BN_bin2bn
```
jonasfj, line 65: This duplicates: https://github.com/google/webcrypto.dart/blob/32ea48b8c4581814f8f751eb2a4ab9bda5551a78/src/symbols.yaml#L45

We should at a minimum put a comment in both places, that they must be kept in sync...

Another option is that make the generator that reads `symbols.yaml` use contents of this file and the other ffigen config file..

lib/src/boringssl/bindings/ffigen.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - nothing
+structs:
+  include:
+    - nothing
+  dependency-only: opaque
+functions:
+  include:
+    - webcrypto_dart_dl_attach_finalizer
+    - webcrypto_dart_dl_initialize
+    - webcrypto_lookup_symbol
```
jonasfj, line 36: Maybe we shouldn't expose the `webcrypto_lookup_symbol` function. We could just do a manual lookup..

A library instance that contains `webcrypto_lookup_symbol` doesn't contain the other ones here.. Not that it matters, it's more of a nit...

jonasfj (approved):

lib/src/boringssl/bindings/ffigen.yaml
dcharkes: https://pub.dev/packages/ffigen/changelog 2.3.0 makes this obsolete.

lib/src/boringssl/bindings/ffigen.yaml
dcharkes: I slightly prefer to keep it, so that we don't have to write the signature manually.

lib/src/boringssl/bindings/ffigen.yaml
jonasfj: Awesome! :D

PR #18, opened 2021-03-15
Remove generic invocation of sizeOf

dcharkes:
Support for invoking `sizeOf<T>` with a generic `T` will be removed in Dart 2.13.

### DIFF
--- a/lib/src/impl_ffi/impl_ffi.rsa_common.dart
+++ b/lib/src/impl_ffi/impl_ffi.rsa_common.dart
@@ -194,8 +194,8 @@ Map<String, dynamic> _exportJwkRsaPrivateOrPublicKey(
     }
 
     // Public key parameters
-    final n = scope.allocate<ffi.Pointer<BIGNUM>>();
-    final e = scope.allocate<ffi.Pointer<BIGNUM>>();
+    final n = scope<ffi.Pointer<BIGNUM>>();
+    final e = scope<ffi.Pointer<BIGNUM>>();
     ssl.RSA_get0_key(rsa, n, e, ffi.nullptr);
 
     if (!isPrivateKey) {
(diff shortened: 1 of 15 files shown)


lib/src/impl_ffi/impl_ffi.utils.dart
```diff
@@ -148,26 +152,28 @@ class _ScopeEntry {
 }
 
 // Utility for tracking and releasing memory.
-class _Scope {
+class _Scope implements Allocator {
```
jonasfj, line 155: Where is this abstraction useful?

lib/src/impl_ffi/impl_ffi.rsa_common.dart
```diff
@@ -208,19 +208,19 @@ Map<String, dynamic> _exportJwkRsaPrivateOrPublicKey(
...
+    final p = scope<ffi.Pointer<BIGNUM>>();
+    final q = scope<ffi.Pointer<BIGNUM>>();
     ssl.RSA_get0_factors(rsa, p, q);
 
-    final dp = scope.allocate<ffi.Pointer<BIGNUM>>();
-    final dq = scope.allocate<ffi.Pointer<BIGNUM>>();
-    final qi = scope.allocate<ffi.Pointer<BIGNUM>>();
+    final dp = scope<ffi.Pointer<BIGNUM>>();
+    final dq = scope<ffi.Pointer<BIGNUM>>();
+    final qi = scope<ffi.Pointer<BIGNUM>>();
```
jonasfj, line 223: Okay, I see this is pretty neat :D

lib/src/impl_ffi/impl_ffi.utils.dart
```diff
@@ -211,35 +217,38 @@ class _Scope {
...
 }
 
-/// Invoke [fn] with [ffi.Pointer<T>] of size count and release the pointer
-/// when [fn] returns.
+/// Invoke [fn] with [p], and release [p] when [fn] returns.
 R _withAllocation<T extends ffi.NativeType, R>(
-  int count,
+  ffi.Pointer<T> p,
   R Function(ffi.Pointer<T>) fn,
 ) {
```
jonasfj, line 232: This suddenly seems a lot less convenient when you have to type the Type being allocated twice..

Can't we keep the old signature? That way it's not so magic that you have yo use `_sslAlloc`.

I think this is a convenience method from before I implemented `_Scope`. Half the places using it should probably be ported to just using `_Scope`. On the other hand I guess it is convenient have this.

Writing: `_withAllocation(1, (ffi.Pointer<CBS> cbs) { ...`  is certainly nice..

jonasfj (approved):
I think this is fine `_withAllocation` should be cleaned up later... Maybe it should be two methods, like:
 * `_withBytes`, for cases that just need `Uint8` arrays,
 * `_withType`, for cases that need a single pointer of type `T`.

I think those are the common cases... but then again, maybe some of them should just be rewritten to use `_Scope` too :D

Anyways, such refactoring can be done later.

lib/src/impl_ffi/impl_ffi.utils.dart
dcharkes: Let me see if I can rewrite the uses to something like:

```dart
  // In a pool multiple resources can be allocated, which will all be freed
  // at the end of the scope.
  using((Pool pool) {
    final p = pool<Int64>(2);
    final p2 = pool<Int64>(2);
    p[0] = 1;
    p[1] = 2;
    MemMove(p2.cast<Void>(), p.cast<Void>(), 2 * sizeOf<Int64>());
    Expect.equals(1, p2[0]);
    Expect.equals(2, p2[1]);
  });
```

https://github.com/dart-lang/sdk/blob/master/samples/ffi/resource_management/pool_sample.dart#L48-L58

The scope is essentially the same as the pool.

jonasfj: Yeah, we could probably have two static methods `_Scope.using` and `_Scope.usingAsync`..

But really, don't sweat it over cleanup..

dcharkes: Tracking in https://github.com/google/webcrypto.dart/issues/19.

Issue #10, opened 2021-03-01
Migrate away from attaching finalizers to pointer objects


jonasfj:
See: https://github.com/dart-lang/sdk/issues/45072

We'll have to refactor code like:
https://github.com/google/webcrypto.dart/blob/6dd34605b16097f2845742e04cce1aa4854b07d8/lib/src/impl_ffi/impl_ffi.utils.dart#L29-L37

As the finalizer will need to be attached to an object that isn't `Pointer`, hence, either we'll probably need to wrap with a `EVP_PKEYResource` or something like that.

mrcsh:
I assume this is the cause of the following error that just showed up after updating to Flutter version 2.3.0-24.1.pre?
```I/flutter (14196): ERROR 11:47:23.000:CryptoBloc._emitError/249:<CryptoBloc>Exception: Assertion failed: "package:webcrypto failed to attached finalizer"
I/flutter (14196): INFO 11:47:23.001:CryptoBloc.emit/168:CryptoState(status:CryptoStatus.error, error=Exception: Assertion failed: "package:webcrypto failed to attached finalizer", keyPair=no) #0      _attachFinalizerEVP_PKEY (package:webcrypto/src/impl_ffi/impl_ffi.utils.dart:39:5)
I/flutter (14196): #1      _createEVP_PKEYwithFinalizer (package:webcrypto/src/impl_ffi/impl_ffi.utils.dart:49:3)
I/flutter (14196): #2      _generateRsaKeyPair (package:webcrypto/src/impl_ffi/impl_ffi.rsa_common.dart:287:21)
I/flutter (14196): #3      rsaOaepPrivateKey_generateKey (package:webcrypto/src/impl_ffi/impl_ffi.rsaoaep.dart:68:16)
I/flutter (14196): #4      RsaOaepPrivateKey.generateKey (package:webcrypto/src/webcrypto/webcrypto.rsaoaep.dart:50:12)
I/flutter (14196): #5      _generateRSAKeyRaw (package:common/bloc/crypto/crypto_bloc.dart:52:28)
I/flutter (14196): #6      _generateRSAKey (package:common/bloc/crypto/crypto_bloc.dart:58:24)
I/flutter (14196): #7      _IsolateConfiguration.apply (package:flutter/src/foundation/_isolates_io.dart:81:34)
I/flutter (14196): #8      _spawn.<anonymous closure> (package:flutter/src/foundation/_isolates_io.dart:88:65)
[...]

PR #25, opened 2022-01-23
Fix finalizers

jonasfj:
Fixes https://github.com/google/webcrypto.dart/issues/10

### DIFF
--- a/lib/src/impl_ffi/impl_ffi.utils.dart
+++ b/lib/src/impl_ffi/impl_ffi.utils.dart
@@ -50,6 +50,72 @@ ffi.Pointer<EVP_PKEY> _createEVP_PKEYwithFinalizer() {
   return key;
 }
 
+@pragma('vm:never-inline')
+Object _finalizerReachabilityFence(Object obj) => obj;
+
+class _EvpPKey {
+  final ffi.Pointer<EVP_PKEY> _pkey;
+
+  factory _EvpPKey() {
+    final _pkey = ssl.EVP_PKEY_new();
+    _checkOp(_pkey.address != 0, fallback: 'allocation failure');
+    return _EvpPKey.wrap(_pkey);
+  }
+
+  _EvpPKey.wrap(this._pkey) {
+    final ret = dl.webcrypto_dart_dl_attach_finalizer(
+      this,
+      _pkey.cast(),
+      ssl.addresses.EVP_PKEY_free.cast(),
+      // We don't really have an estimate of how much space the EVP_PKEY structure
+      // takes up, but if we make it some non-trivial size then hopefully the GC
+      // will prioritize freeing them.
+      4096,
+    );
+    if (ret != 1) {
+      throw AssertionError('package:webcrypto failed to attached finalizer');
+    }
+  }
+
+  T use<T>(T Function(ffi.Pointer<EVP_PKEY> pkey) fn) {
+    try {
+      return fn(_pkey);
+    } finally {
+      _finalizerReachabilityFence(this);
+    }
+  }
+}
+
+extension<T> on T Function(ffi.Pointer<EVP_PKEY>) {
+  T invoke(_EvpPKey key) => key.use((pkey) => this(pkey));
+}
+
+extension<T, A1> on T Function(ffi.Pointer<EVP_PKEY>, A1) {
+  T invoke(_EvpPKey key, A1 arg1) => key.use((pkey) => this(pkey, arg1));
+}
+
+extension<T, A1> on T Function(A1, ffi.Pointer<EVP_PKEY>) {
+  T invoke(A1 arg1, _EvpPKey key) => key.use((pkey) => this(arg1, pkey));
+}
+
+extension<T, A1, A2, A3, A4> on T Function(
+    A1, A2, A3, A4, ffi.Pointer<EVP_PKEY>) {
+  T invoke(
+    A1 arg1,
+    A2 arg2,
+    A3 arg3,
+    A4 arg4,
+    _EvpPKey key,
+  ) =>
+      key.use((pkey) => this(
+            arg1,
+            arg2,
+            arg3,
+            arg4,
+            pkey,
+          ));
+}
+
 /// Throw [OperationError] if [condition] is `false`.
 ///
 /// If [message] is given we use that, otherwise we use error from BoringSSL,
--- a/lib/src/impl_ffi/impl_ffi.utils.dart
+++ b/lib/src/impl_ffi/impl_ffi.utils.dart
@@ -14,54 +14,34 @@
 
 part of impl_ffi;
 
-/// Attach a finalizer for [key], this means that [ssl.EVP_PKEY_free] will
-/// automatically be called with [key] is garbage collected by Dart.
-///
-/// This takes ownership of [key], and the caller **may not** call
-/// [ssl.EVP_PKEY_free].
-///
-/// Callers should be ware that the Dart GC may collect [key] as soon as it
-/// deems the object to not be in use anymore. This can happy at any point when
-/// the VM (optimizer) determines that no code-path passing [key] to an FFI
-/// function can be called again. For this reason, users should take extra care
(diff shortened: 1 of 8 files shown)


lib/src/impl_ffi/impl_ffi.utils.dart
```diff
@@ -14,40 +14,101 @@
...
-/// automatically be called with [key] is garbage collected by Dart.
-///
-/// This takes ownership of [key], and the caller **may not** call
-/// [ssl.EVP_PKEY_free].
+/// Function that can't be inlined to be used for preventing [obj] from being
+/// garbage collected.
+@pragma('vm:never-inline')
+Object _finalizerReachabilityFence(Object obj) {
+  return obj;
+}
```
jonasfj, line 22: @dcharkes is this really the way to do things?

lib/src/impl_ffi/impl_ffi.utils.dart
```diff
@@ -14,40 +14,101 @@
...
+  /// Use the wrapped [EVP_PKEY] in callback [fn].
+  ///
+  /// Note. [fn] is not allowed to return a [Future].
+  T use<T>(T Function(ffi.Pointer<EVP_PKEY> pkey) fn) {
+    try {
+      return fn(_pkey);
+    } finally {
+      _finalizerReachabilityFence(this);
+    }
   }
```
jonasfj, line 69: This is actually find for all use-cases (at-least as far I can understand), the BoringSSL C library has internal reference counting, so all the methods seem as though they take an `EVP_PKEY` and then they hold a reference to it while doing whatever they need to do.

lib/src/impl_ffi/impl_ffi.utils.dart
dcharkes: From Dart 2.17 we should have `implements Finalizable` which will automatically add reachability fences at the end of the syntactic scope for all `Finalizable`s. https://dart-review.googlesource.com/c/sdk/+/227501

lib/src/impl_ffi/impl_ffi.utils.dart
```diff
@@ -14,40 +14,101 @@
...
+    return _EvpPKey.wrap(_pkey);
+  }
+
+  /// Wrap existing [EVP_PKEY], this will attach a finalizer.
+  ///
+  /// After this, the wrapped key may only be used within a callback passed to
+  /// [use]. Otherwise, the garbage collect may be calling the finalizer while
+  /// the key is in use.
+  _EvpPKey.wrap(this._pkey) {
+    final ret = dl.webcrypto_dart_dl_attach_finalizer(
```
dcharkes, line 46: FYI, with Dart 2.17 we will get native finalizers exposed in Dart itself.

* https://github.com/dart-lang/sdk/issues/47777

lib/src/impl_ffi/impl_ffi.utils.dart
```diff
@@ -14,40 +14,101 @@
...
-/// Create an [ssl.EVP_PKEY] with finalizer attached.
-///
-/// See [_attachFinalizerEVP_PKEY] for notes on how the finalizer works.
-ffi.Pointer<EVP_PKEY> _createEVP_PKEYwithFinalizer() {
-  final key = ssl.EVP_PKEY_new();
-  _checkOp(key.address != 0, fallback: 'allocation failure');
-  _attachFinalizerEVP_PKEY(key);
-  return key;
+/// Extension of native function that takes a [EVP_PKEY], making it easy to call
+/// using a wrapped [_EvpPKey].
```
dcharkes, line 73: Instead of this pattern with `use`. Would it be possible to implement them as methods on `_EvpPKey`?

For now, you could add the `reachabilityFence(this)` at the end (and the returns) of those methods. But when `Finalizable` lands, `this` will always be kept alive. See the doc-comment on `Finalizable` in https://dart-review.googlesource.com/c/sdk/+/227501/16/sdk/lib/ffi/native_finalizer.dart.

lib/src/impl_ffi/impl_ffi.utils.dart
jonasfj: Cool, but for this purpose I really don't need it... It could reduce the amount of C code..

lib/src/impl_ffi/impl_ffi.utils.dart
jonasfj: I could, but the `use` method allows me to make extensions like:
```dart
extension<T, A1> on T Function(ffi.Pointer<EVP_PKEY>, A1) {
  /// Invoke this function with unwrapped [key].
  T invoke(_EvpPKey key, A1 arg1) => key.use((pkey) => this(pkey, arg1));
}
```

So then I can write: `ssl.EVP_marshal_private_key.invoke(cbb, _key)`

Instead of having to re-type the signature for `EVP_marshal_private_key` on `_EvpPKey.EVP_marshal_private_key`.

I think I would have to re-type the signature for 10-12 methods... which is kind of annoying, since these methods are generated.

hmm, I guess in some future we can dream about doing something magical with ffigen... hmm... maybe we only dream...

jonasfj: @dcharkes I'm curious, with `Finalizable` what are you going to do if a function requires two arguments that are both wrapped in `Finalizable`?

Do you then have to do double dispatch? And if you don't have a dummy `use` method like mine... will you then have double dispatch for each method.

And what if there is 3 `Finalizable` arguments? do we need triple dispatch? :see_no_evil:

dcharkes: Any variable stays alive (including this, params and normal variables). For more details see the design doc and language proposal.

dcharkes: > I think I would have to re-type the signature for 10-12 methods

Unless all the other things also have Dart types wrapping them, exposing some object oriented API.

dcharkes: Hm, it looks like you have 3 layers of objects: `ffigen`erated objects, then `Finalizable` objects, and then a public API. I was thinking the public API would be `Finalizable`s. But it doesn't look like we have a candidate public API type to make finalizable instead of `_EvpPKey`.

jonasfj: We do `class RsaPssPublicKey extends _EvpPKey`, and do similar for all the key-classes.. I actually did that in one iteration of this, but then I have to pass a constructor-tearoff to helper methods like `_importPkcs8RsaPrivateKey`.

And the whole design got a lot more complicated. If I do that refactoring, I can do a bit more code deduplication.
But it's a larger refactor, so I opted to do that another day. It also wasn't clear that there would be much win to doing it.

But, yes, if `RsaPssPublicKey extends Finalizable`, then I probably won't need to do anything wrapping methods like `EVP_marshal_private_key` at all.

dcharkes: @jonasfj would you like to postpone the refactor? Then I will review the pattern in this CL for now.

jonasfj: I think it'll be easier to review if we do one thing at the time.

This PR changes:
 * All occurences of `ffi.Pointer<EVP_PKEY>` to `_EvpPKey`
 * All methods using `ffi.Pointer<EVP_PKEY>` from `ssl.xxx(key, ...)` to `ssl.xxx.invoke(key, ...)`.
, where `.invoke` is a extension on the function type that calls `_EvpPKey.use`.

That's pretty mechanical... if we migrate to put the finalizer on `RsaPssPublicKey` (and friends), then I think it'll take a lot longer to review.

When we get Dart 2.17, it makes sense to migrate the finalizer to sit on `RsaPssPublicKey` (and friends), and then I won't need the _reachability fence_ and I won't need to do anything magic when calling functions that takes a `ffi.Pointer<EVP_PKEY>`.

dcharkes (approved):
lgtm

PR #26, opened 2022-01-23
Migrate crypto_subtle.dart to use js_util

jonasfj:
@tauu does this also fix your issues?

I think the problem is using things with `@Native` annotation in combination with `@JS` annotations... I'll admit I don't fully understand what is going on, but this seems to work for me.

See also: https://github.com/dart-lang/sdk/issues/44211


lib/src/crypto_subtle.dart
```diff
@@ -116,148 +128,387 @@ class CryptoKeyPair {
...
-  } on FormatException catch (e) {
-    throw UnsupportedError(
-      'exported JsonWebKey is not valid, this could be internal error or '
-      'lacking browser support -- problem: ${e.message}',
-    );
+/// Create [JsonWebKey] from [jsObj].
+JsonWebKey _jsonWebKeyFromJsObj(Object jsObj) {
+  final jwk = JsonWebKey();
+
+  if (hasProperty(jsObj, 'kty')) {
```
sigurdm, line 224: Perhaps make a datastructure for all these. It seems too easy for a mistake to slip through.

```
void setPropIfPresent(String name, void Function(Object?) setter) { 
  if (hasProperty(jsObj, name)) {
    setter(getProperty(jsObj, name);
  }
}

setPropIfPresent('kty', (e) => jwk.kty = e);
setPropIfPresent('use', (e) => jwk.use = e);
...

File: lib/src/crypto_subtle.dart [hunk 2]
```diff
@@ -116,148 +128,387 @@ class CryptoKeyPair {
...
 }
 
-dynamic jsonWebKeytoJs(JsonWebKey k) => js_util.jsify(k.toJson());
+/// Convert [JsonWebKey] to a javascript object.
+Object _jsonWebKeyToJsObj(JsonWebKey jwk) {
+  final jsObj = newObject();
 
-@JS('crypto.getRandomValues')
-external Promise<ByteBuffer> getRandomValues(TypedData array);
+  if (jwk.kty != null) {
```
sigurdm, line 293: Again perhaps you could make a wrapper
```
setPropertyIfPresent(name, val) {
  if (val != null) {
    setProperty(jsObj(name, val));
  }
}
setPropertyIfPresent('kty', jwk.kty);
setPropertyIfPresent('use', jwk.use);
...
```

lib/src/crypto_subtle.dart
jonasfj: A utility like `setPropIfPresent` would certainly make sense

sigurdm (approved):

Issue #22, opened 2021-09-29
Add Flutter Desktop support


dcharkes:
Flutter desktop is becoming more common.

We should add support for flutter desktop.

epoberezkin:
I've managed to run `flutter pub run webcrypto:setup` but `dart test` doesn't work, I can only run `flutter test`...

WiRight:
What about MacOS and Windows?

rokk4:
@WiRight I was able to manually add libcrypto as .dylib and it works.
You can run the compilation on MacOS if you install all the deps.

I also tried it for windows, but I don't know shit about compiling BoringSSL/complicated C++ on Windows I could not make it work. 

@dcharkes I think the biggest problem is setting up the build enviroment on windows, which is a pain in the ...  Would it be a possibility to include precompiled BoringSSL? Or is this not a good idea because the user of the lib would then have to trust the precompiled .dll?

dcharkes:
I believe the build re-exports symbols so that they don't collide. Moreover, the build only uses a subset of BoringSSL which is smaller than the full BoringSSL. So, I don't believe we can use a precompiled dylib from BoringSSL directly.

The Windows setup should be very similar to the Linux setup, the builds are both CMake.
The MacOS setup should be very similar to the iOS setup, the builds are both CocoaPods.

MichaelPriebe:
The error that stops the setup script from working on Windows was fixed in https://github.com/google/webcrypto.dart/commit/851ab9975226769c1f1445d292a77462bf616f39, but even with cmake and gcc, the build fails on Windows.

jonasfj:
This is now resolved in version 0.5.4 which supports Android, iOS, Web, Windows, Linux and Mac.

PR #28, opened 2022-03-07
Add support for Linux desktop

emersion:
References: https://github.com/google/webcrypto.dart/issues/22


linux/webcrypto_plugin.cc
```diff
@@ -0,0 +1,70 @@
+#include "include/webcrypto/webcrypto_plugin.h"
```
jonasfj, line 1: This is just the default template for linux, right?

example/linux/CMakeLists.txt
```diff
@@ -0,0 +1,116 @@
+cmake_minimum_required(VERSION 3.10)
```
jonasfj, line 1: Is `example/linux/` entirely generated by Flutter tool, without manual changes?

emersion: The manual changes are:

- Build `libwebcrypto.so` in `linux/CMakeLists.txt`
- Add it to `webcrypto_bundled_libraries` so that it's getting included in the app bundle

You can test all of this by running the example with `flutter run -d linux`.

linux/webcrypto_plugin.cc
emersion: Just like Android, the plugin stuff generated by `flutter create` is kept intact, but is basically a no-op. It's just used to convince Flutter that Linux is supported in `pubspec.yaml`.

emersion:
Ping

jonasfj (approved):
r+

jonasfj:
Sorry, about the turn around time, I think I just forgot. But this looks great!

emersion:
Thanks!

PR #32, opened 2022-07-28
Fix package resolution and running with `dart test`

serverwentdown:


lib/src/boringssl/lookup/utils.dart
```diff
@@ -102,7 +102,8 @@ Uri? _findDotDartTool() {
 
   // Find script directory
   Uri root;
-  if (Platform.script.isScheme('data')) {
+  if (Platform.script.isScheme('data') ||
+      Platform.script.toFilePath().startsWith("/tmp")) {
```
jonasfj, line 106: Is this a new thing?  If this is for detecting that we're running from tests, then is there no better mechanism we can use?

bin/setup.dart
```diff
@@ -53,7 +53,8 @@ void main() async {
     print('dependency on package:webcrypto is required');
     exit(1);
   }
-  final webcryptoRoot = packageConfigFile.uri.resolve(pkg['rootUri'] ?? '');
+  final webcryptoRoot =
+      packageConfigFile.uri.resolve(pkg['rootUri'] + '/' ?? '');
```
jonasfj, line 57: Not required on my end, but if we want to go down this route, then we should make it conditional to avoid duplicate slashes.

jonasfj:
Thanks for the PR, I think I fixed this slightly differently in https://github.com/google/webcrypto.dart/pull/35

PR #35, opened 2022-08-18
Fix logic for finding libwebcrypto.so in tests

jonasfj:


lib/src/boringssl/lookup/utils.dart
```diff
@@ -117,5 +110,18 @@ Uri? _findDotDartTool() {
       return root.resolve('.dart_tool/');
     }
   } while (root != (root = root.resolve('..')));
+
+  // If travesing from script directory didn't work, we can look starting from
```
sigurdm, line 114:
```suggestion
  // If traversing from script directory didn't work, we can look starting from
```

lib/src/boringssl/lookup/utils.dart
```diff
@@ -101,14 +101,7 @@ Uri? _findDotDartTool() {
   //       `.dart_tool/package_config.json` file.
```
sigurdm, line 101: This comment is misleading. dart:isolate is available, but the function for getting the packageConfig is async...

sigurdm (approved):
LGTM

Issue #20, opened 2021-03-24
webcrypto:setup: CMake Error: The source directory does not exist


xvrh:
I try to use this library in a small dart script (inside a flutter project), and I got:
```
Unsupported operation: package:webcrypto cannot be used from Dart or `pub run test` unless `flutter pub run webcrypto:setup` has been run for the current root project.
```

So I try:
```
> flutter pub run webcrypto:setup   
Building with assumed project root in:
/Users/xavier/projects/app/
Using package:webcrypto from /Users/xavier/.pub-cache/hosted/pub.dartlang.org/webcrypto-0.5.2
Generating build system with cmake
CMake Error: The source directory "/Users/xavier/.pub-cache/hosted/pub.dartlang.org/src" does not exist.
Specify --help for usage, or press the help button on the CMake GUI.
Generating with cmake failed, ensure you have dependencies!
```

Webcrypto version: 0.5.2

```
> flutter --version
Flutter 2.0.3 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 4d7946a68d (6 days ago) • 2021-03-18 17:24:33 -0700
Engine • revision 3459eb2436
Tools • Dart 2.12.2

> cmake --version
cmake version 3.16.1
```

xvrh:
It seems the code here https://github.com/google/webcrypto.dart/blob/master/bin/setup.dart#L64 is incorrect

The `rootUri` can be something like`file:///Users/xavier/.pub-cache/hosted/pub.dartlang.org/webcrypto-0.5.2`

And 
```dart
Uri.parse('file:///Users/xavier/.pub-cache/hosted/pub.dartlang.org/webcrypto-0.5.2').resolve('src');
```
gives `file:///Users/xavier/.pub-cache/hosted/pub.dartlang.org/src`

prijindal:
To forcefully run this, in your file:

~/.pub-cache/hosted/pub.dartlang.org/webcrypto-0.5.2/bin/setup.dart

on line 64, replace
```
webcryptoRoot.resolve('src').toFilePath(),
```

 with 
```
webcryptoRoot.resolve('webcrypto-0.5.2/src').toFilePath(),
```

It is a temporary fix

jonasfj:
Thanks for the report, I wonder how this worked locally on my machine.

prijindal:
For a one liner fix with bash:
```bash
[...]

PR #40, opened 2022-09-01
Fix: setup.dart failed to run under Linux

maltemedocs:
`webcryptoRoot.resolve('src')` did not get evaluated to `$PUB_CACHE/webcrypto-x.y.z/src`, but just to `$PUB_CACHE/src`. This broke `flutter pub run webcrypto:setup` at the very least on Linux, but probably on more systems as well. Replacing the URI.resolve approach with a simple string concatenation fixes the bug.

Fixes https://github.com/google/webcrypto.dart/issues/20

### DIFF
--- a/bin/setup.dart
+++ b/bin/setup.dart
@@ -63,7 +63,8 @@ void main() async {
     'cmake',
     [
       '-S',
-      webcryptoRoot.resolve('src').toFilePath(),
+      Directory(webcryptoRoot.toFilePath() + Platform.pathSeparator + 'src')
+          .path,
       '-B',
       root.resolve('.dart_tool/webcrypto').toFilePath(),
     ],
--- a/bin/setup.dart
+++ b/bin/setup.dart
@@ -63,7 +63,7 @@ void main() async {
     'cmake',
     [
       '-S',
-      Directory(webcryptoRoot.toFilePath() + Platform.pathSeparator + 'src')
+      Directory(joinPaths(webcryptoRoot.toFilePath(), 'src'))
           .path,
       '-B',
       root.resolve('.dart_tool/webcrypto').toFilePath(),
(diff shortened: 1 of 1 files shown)


maltemedocs:
@jonasfj Worth it to go through the CLA process for such a simple fix, and would it even be merged?

jonasfj (approved):

jonasfj:
@maltemedocs, I think this is safe to be merged.

bin/setup.dart
```diff
@@ -63,7 +63,8 @@ void main() async {
     'cmake',
     [
       '-S',
-      webcryptoRoot.resolve('src').toFilePath(),
+      Directory(webcryptoRoot.toFilePath() + Platform.pathSeparator + 'src')
+          .path,
```
jonasfj, line 67: Should we do something like:
```dart
/// Join paths without duplicate path separators.
String joinPaths(String prefix, String suffix) {
  if (!prefix.endsWith(Platform.pathSeparator)) {
    prefix += Platform.pathSeparator;
  }
  return prefix + suffix;
}
```

maltemedocs: In the systems I'm aware of, directory1//directory2 == directory1/directory2, so it should not be necessary. But it would not hurt ofc if you feel this is cleaner.

jonasfj: It would be nice to be clean. Mostly because it might affect other things down the line (probably won't).

jonasfj:
@maltemedocs still need the CLA though. Or I can just land something similar to this :D

bin/setup.dart
maltemedocs: I added it in https://github.com/google/webcrypto.dart/pull/40/commits/4754ccc23fac6470f58207e16f0e5da9df07c6c3 (your approach worked perfectly in my test).

maltemedocs:
Ok, I'll let my boss know and try to get this done. If it takes too long I'm okay with you landing something similar :)

bin/setup.dart
```diff
@@ -98,3 +99,11 @@ void main() async {
   print('This is only necessary for using package:webcrypto in unit tests');
   print('and scripts, not for usage in applications.');
 }
+
+/// Join paths without duplicate path separators.
+String joinPaths(String prefix, String suffix) {
```
jonasfj, line 104: nit:
```suggestion
String _joinPaths(String prefix, String suffix) {
```

jonasfj (approved):
Let's merge once CLA passes.

maltemedocs:
CLA is signed! :)

jonasfj:
Thanks!

PR #41, opened 2022-09-22
Use native finalizer

jonasfj:
@dcharkes, mostly curious if the first commit is correct... the others are just removing stuff.

I probably still have to figure out what's wrong with the macos tests. I'm guessing I just forgot something trivial.


lib/src/impl_ffi/impl_ffi.utils.dart
```diff
@@ -43,30 +40,13 @@ class _EvpPKey {
   /// [use]. Otherwise, the garbage collect may be calling the finalizer while
   /// the key is in use.
   _EvpPKey.wrap(this._pkey) {
```
dcharkes, line 42: Nit: from a resource perspective, I'd make `wrap` the unnamed constructor. And `new` or `allocate` a factory.

lib/src/impl_ffi/impl_ffi.utils.dart
```diff
@@ -14,27 +14,24 @@
...
   factory _EvpPKey() {
-    final _pkey = ssl.EVP_PKEY_new();
-    _checkOp(_pkey.address != 0, fallback: 'allocation failure');
-    return _EvpPKey.wrap(_pkey);
+    final pkey = ssl.EVP_PKEY_new();
+    _checkOp(pkey.address != 0, fallback: 'allocation failure');
+    return _EvpPKey.wrap(pkey);
   }
 
   /// Wrap existing [EVP_PKEY], this will attach a finalizer.
```
dcharkes, line 37: Do you ever want to early-free a key?

Usually native resources have a `_released`/`_deleted` internal boolean that is being check before every method and a `release`/`delete` method that removes the finalizer and calls the native "free" early.
(Even better if that works also with Arena's: https://github.com/dart-lang/sdk/issues/49906 )

dcharkes (commented):
commit adding native finalizers lgtm

dcharkes (approved):
lgtm rest

bumped the referred ffigen issue

jonasfj:
Couldn't help myself to not just regenerate bindings, and accidentally added it to this PR :see_no_evil:  but I did split into nice small commits :D

Anyways, I probably have to work a bit to make tests pass again, I have no idea why they broken on Mac OS, probably some random stupid thing when I started messing with podspec...

lib/src/impl_ffi/impl_ffi.utils.dart
jonasfj: Not really, I don't think I have any use-cases where I release early.

lib/src/third_party/boringssl/ffigen.yaml
```diff
@@ -1,17 +1,34 @@
 name: BoringSsl
 description: 'Bindings to BoringSSL.'
+language: c
 output: 'lib/src/third_party/boringssl/generated_bindings.dart'
 headers:
   entry-points:
-    - 'third_party/boringssl/src/include/openssl/*.h'
+    # bytestring.h MUST be the first header, otherwise structs are just left
+    # as opaque.
+    - 'third_party/boringssl/src/include/openssl/bytestring.h'
```
jonasfj, line 9: @dcharkes, @mannprerak2, I have a scenario here where `base.h` forward declares `CBS` as:

```c
// from base.h
typedef struct cbs_st CBS;
```

But in `bytestring.h` we have the full `cbs_st` declaration:
```c
// from bytestring.h
struct cbs_st {
  const uint8_t *data;
  size_t len;
};
```

If `base.h` is read first by ffigen, then I get:
```dart
class cbs_st extends ffi.Opaque {}
```

But if `bytestring.h` is read first, then I get:
```dart
class cbs_st extends ffi.Struct {
  external ffi.Pointer<ffi.Uint8> data;

  @ffi.Size()
  external int len;
}
```
(which is what I need in order to be able to allocate and instance of the struct)

--------------------------------

All the other headers include `base.h` (or at-least many of them do). So my options here are:
 * Put `bytestring.h` at the top of the list.
 * Create my own `amalgamation.h` which includes all the headers I need.

I'm wondering if this is working as intended, or if it's a bug? I suspect this pattern is common in C-libraries.

Issue #39, opened 2022-08-22
update ffigen


jonasfj:
Running:
```
flutter pub run ffigen --config=lib/src/boringssl/bindings/ffigen.yaml
flutter pub run ffigen --config=lib/src/third_party/boringssl/ffigen.yaml
```
causes things like:
```
[WARNING]: Resolved name conflict: Declaration 'RSA_set0_key' and has been renamed to 'RSA_set0_key1'.
[WARNING]: Resolved name conflict: Declaration 'RSA_set0_key' and has been renamed to 'RSA_set0_key2'.
[WARNING]: Resolved name conflict: Declaration 'RSA_set0_key' and has been renamed to 'RSA_set0_key3'.
[WARNING]: Resolved name conflict: Declaration 'RSA_set0_key' and has been renamed to 'RSA_set0_key4'.
[WARNING]: Resolved name conflict: Declaration 'RSA_set0_key' and has been renamed to 'RSA_set0_key5'.
[WARNING]: Resolved name conflict: Declaration 'RSA_set0_key' and has been renamed to 'RSA_set0_key6'.
```
where multiple methods for the same thing is generated.

@dcharkes, any ideas? (this is upgrading from `ffigen` version 4 to 6)

dcharkes:
Sounds like:

* https://github.com/dart-lang/ffigen/issues/384

dcharkes:
Please file on ffigen repo.

jonasfj:
I think I have resolved this by tweaking the configuration. Still need to brush off a few things...

PR #47, opened 2022-10-17
Regenerated ffi bindings

jonasfj:
If you have any suggestions for better config I'm all ears, but this did reduce the number of useless bindings a bit..

I think I I got rid of pthreads and stdio :rofl: 

However,
```yaml
structs:
  include:
    - cbs_st
    - cbb_st
  dependency-only: opaque
```
Doesn't seem to make everything other than `cbs_st` and `cbb_st` opaque... so maybe I'm configuring it wrong.

In any case, it's much less overhead then having pthreads and stdio :see_no_evil: and none of this should have runtime implications as the treeshaker will hopefully eat all of it.

Fixes https://github.com/google/webcrypto.dart/issues/39

### DIFF
--- a/lib/src/third_party/boringssl/generated_bindings.dart
+++ b/lib/src/third_party/boringssl/generated_bindings.dart
@@ -2969,67 +2969,6 @@ class BoringSsl {
       int Function(ffi.Pointer<RSA>, ffi.Pointer<BIGNUM>, ffi.Pointer<BIGNUM>,
           ffi.Pointer<BIGNUM>)>();
 
-  late final ffi.Pointer<ffi.Int> ___daylight = _lookup<ffi.Int>('__daylight');
-
-  int get __daylight => ___daylight.value;
-
-  set __daylight(int value) => ___daylight.value = value;
-
-  late final ffi.Pointer<ffi.Long> ___timezone =
-      _lookup<ffi.Long>('__timezone');
-
-  int get __timezone => ___timezone.value;
-
-  set __timezone(int value) => ___timezone.value = value;
-
-  late final ffi.Pointer<ffi.Pointer<ffi.Pointer<ffi.Char>>> ___tzname =
-      _lookup<ffi.Pointer<ffi.Pointer<ffi.Char>>>('__tzname');
-
-  ffi.Pointer<ffi.Pointer<ffi.Char>> get __tzname => ___tzname.value;
-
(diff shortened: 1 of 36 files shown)


dcharkes:
> Doesn't seem to make everything other than `cbs_st` and `cbb_st` opaque... so maybe I'm configuring it wrong.

What types did you expect to get opaque as well?

lib/src/third_party/boringssl/generated_bindings.dart
```diff
@@ -3539,56 +3211,59 @@ class crypto_mutex_st extends ffi.Union {
...
+  external ffi.Pointer<ffi.Void> md_data;
+
+  /// pctx is an opaque (at this layer) pointer to additional context that
+  /// EVP_PKEY functions may store in this object.
+  external ffi.Pointer<EVP_PKEY_CTX> pctx;
+
+  /// pctx_ops, if not NULL, points to a vtable that contains functions to
+  /// manipulate |pctx|.
+  external ffi.Pointer<evp_md_pctx_ops> pctx_ops;
+}
```
jonasfj, line 3251: This became non-opaque... I'm not sure why?

Since I think I asked for dependencies to be opaque, and only listed `cbs_st` and `cbb_st` in `includes`.
But it seems to make have everything included by dependency as non-opaque.

It's not a huge issue.

dcharkes: cc @mannprerak2

mannprerak2: Is it possible that this struct is being referred to by Value somewhere? Since opaque only works with pointer references. (I do see its usage by value via a typedef reference) 

Maybe compare the before and after logs for v 7.0.0 and v 7.1.0
For the older version, logs should contain warnings like - "skipped <declaration> due to incomplete member reference"
But since 7.1.0 they are now complete (as we can now find definitions across entry-points).

mannprerak2: Another case I found is when something like this happens - 

```C
void func(struct A at, struct OpaqueByIntention to);
```
Suppose this is the only place where `struct A` is used by value, then generated A will not be Opaque, even though `func` (that used A by value) is skipped due to `OpaqueByIntention` pass by value. 
This happens because of the way we parse our code, if the config is set to `dependency-only: opaque` we immediately parse the dependencies if the struct at any point has been referred by value.

PR #74, opened 2024-01-10
Enable desktop macos testing

jonasfj:

PR #77, opened 2024-01-10
Enable mac tests

jsiedentop:
When the example app is not in the background during the test, all tests are green on my local mac. Furthermore I was able to successfully execute the *tests.yml* workflow on Github Actions without any relevant change. I hope that this can be reproduced on the main branch and #74 is fixed.

image

Finally, I have fixed a warning and adjusted the readme regarding the supported platforms.

### DIFF
--- a/lib/src/webcrypto/webcrypto.hmac.dart
+++ b/lib/src/webcrypto/webcrypto.hmac.dart
@@ -197,7 +197,7 @@ abstract class HmacSecretKey {
   /// timing attacks. To validate signatures use [verifyBytes] or [verifyStream]
   /// instead, these methods computes a signature and does a
   /// fixed-time comparison.
-  /// {@template}
+  /// {@endtemplate}
   Future<Uint8List> signBytes(List<int> data);
 
   /// Compute an HMAC signature of given [data] stream.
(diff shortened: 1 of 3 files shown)


lib/src/webcrypto/webcrypto.hmac.dart
```diff
@@ -197,7 +197,7 @@ abstract class HmacSecretKey {
   /// timing attacks. To validate signatures use [verifyBytes] or [verifyStream]
   /// instead, these methods computes a signature and does a
   /// fixed-time comparison.
-  /// {@template}
+  /// {@endtemplate}
```
jonasfj, line 200: For future reference please submit unrelated changes as individual PRs.

That makes it much easier to review and land them, and reduces the risk of conflicts, etc.

Small and fast is always preferable.

jonasfj (approved):

jonasfj:
Hmm, maybe my run just timed out due to bad luck  :see_no_evil:  or maybe you're just better at enabling tests than I am :rofl:

Either way, let's merge it and ship this!

jsiedentop:
Normally I delete commented out code. Maybe that left an impression, because if it hadn't worked, I would have started deleting it 🤫

PR #86, opened 2024-02-25
Migrate lib/src/crypto_subtle.dart to use `dart:js_interop`

### LINKED ISSUE #83: Migrate lib/src/crypto_subtle.dart to use `dart:js_interop`
Using `dart:js_interop` or `package:web` instead of other libraries will make this wasm-ready.

Migrating https://github.com/google/webcrypto.dart/blob/master/lib/src/crypto_subtle.dart should be possible.

koji-1009:
Migrate web crypto api  code to `dart:js_interop`.

I have added `JS` to class names, like `JSWindow`, to distinguish it from classes that are not related to `dart:js_interop`. But if this is unnecessary, please point it out in review comments.

close #83

### DIFF
--- a/lib/src/crypto_subtle.dart
+++ b/lib/src/crypto_subtle.dart
@@ -113,7 +113,7 @@ extension type JSSubtleCrypto(JSObject _) implements JSObject {
 
   /// https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/generateKey
   @JS('generateKey')
-  external JSPromise<JSCryptoKey> generateCreptoKey(
+  external JSPromise<JSCryptoKey> generateCryptoKey(
     JSAny algorithm,
     bool extractable,
     JSArray<JSString> keyUsages,
@@ -121,7 +121,7 @@ extension type JSSubtleCrypto(JSObject _) implements JSObject {
 
   /// https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/generateKey
   @JS('generateKey')
-  external JSPromise<JSCryptoKeyPair> generateCreptoKeyPair(
+  external JSPromise<JSCryptoKeyPair> generateCryptoKeyPair(
     JSAny algorithm,
     bool extractable,
     JSArray<JSString> keyUsages,
@@ -422,7 +422,7 @@ Future<JSCryptoKey> generateKey(
   List<String> usages,
 ) async {
   final value = await window.crypto.subtle
-      .generateCreptoKey(
+      .generateCryptoKey(
         algorithm.toJS,
         extractable,
         usages.toJS,
@@ -438,7 +438,7 @@ Future<JSCryptoKeyPair> generateKeyPair(
   List<String> usages,
 ) async {
   final value = await window.crypto.subtle
-      .generateCreptoKeyPair(
+      .generateCryptoKeyPair(
         algorithm.toJS,
         extractable,
         usages.toJS,
--- a/lib/src/crypto_subtle.dart
+++ b/lib/src/crypto_subtle.dart
@@ -19,6 +19,8 @@ library common;
 import 'dart:js_interop';
 import 'dart:typed_data';
 
+import 'package:meta/meta.dart';
+
 import 'jsonwebkey.dart' show JsonWebKey, RsaOtherPrimesInfo;
 
 export 'jsonwebkey.dart' show JsonWebKey;
@@ -554,15 +556,17 @@ Future<ByteBuffer> deriveBits(
   return value.toDart;
 }
 
-extension on List<String> {
+extension ListExtension on List<String> {
+  @visibleForTesting
   JSArray<JSString> get toJS => <JSString>[
         for (final value in this) value.toJS,
       ].toJS;
 }
 
-extension on Algorithm {
+extension AlgorithmExtension on Algorithm {
   /// Create JSON from [Algorithm].
   /// To avoid null properties for keys, eliminate keys whose value is null.
+  @visibleForTesting
   JSAny get toJS {
     final json = <String, Object>{};
     final name_ = name;
--- a/lib/src/crypto_subtle.dart
+++ b/lib/src/crypto_subtle.dart
@@ -171,11 +171,12 @@ extension type JSCryptoKey(JSObject _) implements JSObject {
   ///  * `'private'`
   ///  * `'public'`
   ///  * `'secret'`
+  ///
   /// https://developer.mozilla.org/en-US/docs/Web/API/CryptoKey/type
   external String get type;
 
   /// True, if this key can be extracted.
-
+  ///
   /// https://developer.mozilla.org/en-US/docs/Web/API/CryptoKey/extractable
   external bool get extractable;
 
@@ -188,6 +189,7 @@ extension type JSCryptoKey(JSObject _) implements JSObject {
   ///  * `'deriveBits'`,
   ///  * `'wrapKey'`,
   ///  * `'unwrapKey'`.
+  ///
   /// https://developer.mozilla.org/en-US/docs/Web/API/CryptoKey/usages
   external JSArray<JSString> get usages;
 }
--- a/lib/src/crypto_subtle.dart
+++ b/lib/src/crypto_subtle.dart
@@ -332,23 +332,23 @@ extension type JSRsaOtherPrimesInfo(JSObject _) implements JSObject {
 
 TypedData getRandomValues(TypedData array) {
   if (array is Uint8List) {
-    final result = window.crypto.getRandomValues(array.toJS) as JSUint8Array;
-    return result.toDart;
+    window.crypto.getRandomValues(array.toJS);
+    return array;
   } else if (array is Uint16List) {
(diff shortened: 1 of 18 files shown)


lib/src/testing/utils/testrunner.dart
```diff
@@ -809,7 +809,7 @@ void _runTests<PrivateKey, PublicKey>(
       signature = c.signature;
     } else {
       test('create signature', () async {
-        signature = await r._signBytes!(
+        signature = await r._signBytes(
```
jonasfj, line 812: Why these changes? Is it perhaps because we updated the language version when we bumped the SDK constraint? (Just guessing that maybe this had an impact on inference for nullability)

If so, can we consider bumping the SDK constraint in a differnet PR?

lib/src/crypto_subtle.dart
```diff
@@ -73,22 +42,142 @@ Uint8List bigIntToUint8ListBigInteger(BigInt integer) {
...
+  /// https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/deriveBits
+  external JSPromise<JSArrayBuffer> deriveBits(
+    JSAny algorithm,
+    JSCryptoKey baseKey,
+    int length,
+  );
+
+  /// https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/generateKey
+  @JS('generateKey')
+  external JSPromise<JSCryptoKey> generateCreptoKey(
```
jonasfj, line 116:
```suggestion
  external JSPromise<JSCryptoKey> generateCryptoKey(
```

typo?

lib/src/crypto_subtle.dart
```diff
@@ -73,22 +42,142 @@ Uint8List bigIntToUint8ListBigInteger(BigInt integer) {
...
+    JSAny algorithm,
+    bool extractable,
+    JSArray<JSString> keyUsages,
+  );
+
+  /// https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/generateKey
+  @JS('generateKey')
+  external JSPromise<JSCryptoKeyPair> generateCreptoKeyPair(
```
jonasfj, line 124: typo?

lib/src/crypto_subtle.dart
```diff
@@ -73,22 +42,142 @@ Uint8List bigIntToUint8ListBigInteger(BigInt integer) {
...
   ///  * `'private'`
   ///  * `'public'`
   ///  * `'secret'`
-  String get type => getProperty(_jsObj, 'type');
+  /// https://developer.mozilla.org/en-US/docs/Web/API/CryptoKey/type
+  external String get type;
 
   /// True, if this key can be extracted.
-  bool get extractable => getProperty(_jsObj, 'extractable');
+
```
jonasfj, line 178:
```suggestion
///
```
missing empty documentation comment line?

lib/src/impl_js/impl_js.random.dart
```diff
@@ -17,7 +17,7 @@ part of impl_js;
 void fillRandomBytes(TypedData destination) {
   try {
     subtle.getRandomValues(destination);
-  } on subtle.DomException catch (e) {
+  } on subtle.JSDomException catch (e) {
```
jonasfj, line 20: Does this work?

lib/src/impl_js/impl_js.utils.dart
```diff
@@ -107,7 +107,7 @@ Future<T> _handleDomException<T>(
 }) async {
   try {
     return await fn();
-  } on subtle.DomException catch (e) {
+  } on subtle.JSDomException catch (e) {
```
jonasfj, line 110: Will this work?

lib/src/crypto_subtle.dart
```diff
@@ -245,242 +300,375 @@ class Algorithm {
...
+  external String get t;
 }
 
-/// Convert a [list] a Javascript object.
-Object _stringListToJsObj(List<String> list) => callConstructor(
-      _array,
-      list,
+TypedData getRandomValues(TypedData array) {
+  if (array is Uint8List) {
+    final result = window.crypto.getRandomValues(array.toJS) as JSUint8Array;
```
jonasfj, line 333: We could perhaps make this simpler by throwing away the return value.

We can just `return array`, since that's also what the API returns:
https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues#return_value

lib/src/impl_js/impl_js.random.dart
jonasfj: Can we make some unit tests for `impl_js`.

Like:
```js
crypto.getRandomValues(new Uint8Array(1000000))
// throws:
// Uncaught DOMException: Failed to execute 'getRandomValues' on 'Crypto': The ArrayBufferView's byte length (1000000) exceeds the number of bytes of entropy available via this API (65536).
```

at-least in chrome.


-----------------


Are there scenarios where we could get `TypeError` or other such things when calling into JS, and would handle them correctly?

jonasfj (commented):
Mostly just nits, a few dumb questions about exception handling.

But overall this looks good.

lib/src/testing/utils/testrunner.dart
koji-1009: As you may have guessed, this is what lint pointed out because of the newer version of dart. The following PR separates the Flutter and Dart version upgrades and `dart fix --apply`.

https://github.com/google/webcrypto.dart/pull/87

lib/src/crypto_subtle.dart
koji-1009: Sorry, typo, I'll fix the two places together.

lib/src/crypto_subtle.dart
koji-1009: Thank you! 
I will give it a try!

lib/src/crypto_subtle.dart
koji-1009: It's typo. I fixed. [e9c8900](https://github.com/google/webcrypto.dart/pull/86/commits/e9c890031c926e235bed935a9bf590a3bde4e7e5)

lib/src/crypto_subtle.dart
koji-1009: It's typo. I fixed. [e9c8900](https://github.com/google/webcrypto.dart/pull/86/commits/e9c890031c926e235bed935a9bf590a3bde4e7e5)

lib/src/crypto_subtle.dart
koji-1009: I added empty lines. [55af27e](https://github.com/google/webcrypto.dart/pull/86/commits/55af27e9403b77c5437f947ee9158f7b7805838a)

lib/src/impl_js/impl_js.random.dart
koji-1009: I added a test file for each browser since the error is expressed differently in chrome and firefox. And, I was also able to add a scenario where DOMException is try-catch (a bit weird).
Does this look good?

* [chrome](https://github.com/google/webcrypto.dart/pull/86/files#diff-627e300589c395ca34cb9c192df9b8ea491ed4b16f67cdeaa9b90cd50a4cd441)
* [firefox](https://github.com/google/webcrypto.dart/pull/86/files#diff-8f2af7a8565ec85972baa2d7444e5765368cc42f8905f46df6cac4c6ab2b942d)

ref https://pub.dev/packages/test#restricting-tests-to-certain-platforms

lib/src/impl_js/impl_js.utils.dart
koji-1009: Strange as it may seem, try-catch is possible..

lib/src/impl_js/impl_js.random.dart
jonasfj: Hmm, I guess this works :rocket:

I'm just getting used to that everything returned or thrown by these APIs is in fact a `JSObject`. And any instance of `JSObject` is also an instance of `subtle.JSDomException`.

So any JS exception thrown will be a `subtle.JSDomException`. But that's probably fine, since we always read the `.name` property in order to decide what to do.

test/browser_chrome_test.dart
```diff
@@ -0,0 +1,186 @@
+@TestOn('chrome')
```
jonasfj, line 1: I imagine that the two files for Firefox and Chrome are quite similar.

Can we make this a single file.

Perhaps we can make `detectedRuntime` to handle different browsers differently.

If we only have a single file it's easier to spot the difference.

jonasfj (approved):
I think this is good enough to ship!

r+

If you want to tweak the test things, feel free. Otherwise, I'll land this, and be happy to accept PRs improving tests.

Writing tests for `crypto_subtle.dart` actually makes sense. But it's a non-trivial amount of work, and certainly not something we should block this PR on.

jonasfj:
I'll get this out in the next release, feel free to file a PR preparing a release (bump version and update changelog), if you want it sooner.

koji-1009:
@jonasfj 
Thanks for the review and advice!
I will work on improving the test over the weekend.

PR #88, opened 2024-03-02
Unify browser test files and add safari tests

koji-1009:
ref https://github.com/google/webcrypto.dart/pull/86#pullrequestreview-1906593118

As a test for `crypto_subtle.dart`, I have combined them into one test file.
By specifying `test.testOn`, different expectations are expressed in chrome, firefox and safari.

https://pub.dev/documentation/test/latest/test/test.html

I checked and got an `OperationError` when publicExponent is `Uint8List.fromList([0x03])` in macOS + safari. I added a test case to `crypto_subtle_dart.dart`, but it seems difficult to include macOS + safari in the `webcrypto_test.dart` test target.
And on macOS, the firefox test cannot be run due to the following issue.

https://github.com/dart-lang/test/issues/2194

Due to these reasons, I have not made any changes to the workflows.


test/crypto_subtle_test.dart
```diff
@@ -0,0 +1,537 @@
...
+          isA<subtle.JSDomException>()
+              .having(
+                (e) => e.name,
+                'name',
+                'TypeMismatchError',
+              )
+              .having(
+                (e) => e.message,
+                'message',
+                'The type of an object is incompatible with the expected type of the parameter associated to the object',
```
jonasfj, line 194: I do wonder, if perhaps we'd be better off only testing `e.name` and not testing `e.message`.

It's not really part of the specification what the message should be, and I don't think we'll benefit from having test cases that break whenever Chrome or Firefox is updated and starts producing better error messages.

But ideally, the kind of exception they throw `e.name` should be somewhat stable.

jonasfj (approved):

test/crypto_subtle_test.dart
koji-1009: @jonasfj 
Hi, I pushed [ba1ff2e](https://github.com/google/webcrypto.dart/pull/88/commits/ba1ff2ec9df2af55bf1af0a7cbf945ee0c0aa1d0) to simplify test cases. Could you review it again? Thanks.

test/crypto_subtle_test.dart
```diff
@@ -0,0 +1,311 @@
...
+              const subtle.Algorithm().toJS,
+              false,
+              ['encrypt', 'decrypt'].toJS,
+            )
+            .toDart,
+        throwsA(
+          isA<subtle.JSDomException>().having(
+            (e) => e.name,
+            'name',
+            'TypeError',
```
jonasfj, line 166: nit:
```suggestion
            // Chrome / Safari throw TypeError
            // Firefox throws SyntaxError
            anyOf('TypeError', 'SyntaxError'),
```
Testing that it's either a `TypeError` or `SyntaxError` would also be fine.

jonasfj (approved):

test/crypto_subtle_test.dart
koji-1009: Thank you! I thought it was good!

jonasfj:
Thanks!

This is awesome.

hmm, sad to see the iOS integration tests failing so frequently. I'll file a bug, someday it might be worth fixing.

PR #89, opened 2024-03-02
chore: Update actions

koji-1009:
Action updated because Node 16 is now deprecated.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -13,7 +13,7 @@ jobs:
     permissions:
       id-token: write
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


jonasfj (approved):

jonasfj:
I suspect the iOS failure is an intermittent error, I'll retry it once or twice. Probably we should work to figure it out why it fail sometimes (or rather why it only succeeds sometimes :see_no_evil:)

jonasfj:
Thanks for cleaning this up. I see the iOS finally passed :see_no_evil: 

It seems like it's doing everything right, but somehow it gets stuck and the test has a timeout.
Not sure if it's waiting somewhere in [integration_test/webcrypto_test.dart](https://github.com/google/webcrypto.dart/blob/master/example/integration_test/webcrypto_test.dart), or how this happens.

Issue #55, opened 2022-11-02
Extend TestRunner with support for exception and error cases


jonasfj:
Maybe we could somehow have support for test vectors where the importing the key throws an exception/error, and where we assert certain errors/exceptions.

This might be different from `_TestCase`, but to make it fast maybe it can still be part of `TestRunner`

PR #100, opened 2024-05-01
Documentation for most AES-GCM members

jonasfj:


sigurdm:
Does this documentation come from somewhere, or you wrote it all from scratch?

lib/src/webcrypto/webcrypto.aesgcm.dart
```diff
@@ -14,40 +14,263 @@
...
 
-  // TODO: Document that this does not provide a streaming interface because
-  //       access to the decrypted bytes before verification of the
-  //       authentication tag defeats the purpose of authenticated-encryption.
+  /// Encrypt [data] with this [AesCbcSecretKey] using AES in
+  /// _Galois/Counter Mode_ (GCM-mode), as specified in [NIST SP 800-38D][1].
+  ///
+  /// This operation requires an _initalization vector_ [iv]. The [iv]
+  /// needs not be secret, but it must unique for each invocation.
+  /// In particular the same (key, [iv]) pair must be used more than once.
```
sigurdm, line 134:
```suggestion
  /// In particular the same (key, [iv]) pair must **not** be used more than once.
```

lib/src/webcrypto/webcrypto.aesgcm.dart
```diff
@@ -14,40 +14,263 @@
...
     int? tagLength = 128,
   });
 
+  /// Export [AesGcmSecretKey] as raw bytes.
+  ///
+  /// This returns raw bytes making up the secret key.
+  ///
+  /// **Example**
+  /// ```dart
+  /// import 'package:webcrypto/webcrypto.dart';
```
sigurdm, line 237: It nags me that the example is both top-level code, and statement-level code...
Ideally it would somehow be actual runnable code by some convention.

lib/src/webcrypto/webcrypto.aesgcm.dart
```diff
@@ -14,40 +14,263 @@
...
+  /// ```
+  /// {@endtemplate}
+  ///
+  /// {@template AesGcmSecretKey-remark:no-stream-api}
+  /// **Remark** this package does not offer a streaming API for
+  /// encryption / decryption using AES-GCM, because reading deciphered
+  /// plaintext prior to complete verification of the tag breaks the
+  /// authenticity assurances. Specifically, until the entire message is
+  /// decrypted it is not possible to know if it is authentic, which would
+  /// defeat the purpose of _authenticated encryption_.
```
sigurdm, line 210: Theoretically there could be a streaming API for computational reasons (ie. you don't have to keep everything in memory at once), ending up with the result  being thrown away at the end.

lib/src/webcrypto/webcrypto.aesgcm.dart
```diff
@@ -14,40 +14,263 @@
...
+  /// Map<String, dynamic> keyData = await key.exportJsonWebKey();
+  /// print(jsonEncode(keyData));
+  /// ```
+  ///
+  /// [1]: https://tools.ietf.org/html/rfc7517
   static Future<AesGcmSecretKey> importJsonWebKey(Map<String, dynamic> jwk) {
     return impl.aesGcm_importJsonWebKey(jwk);
   }
 
+  /// Generate random [AesGcmSecretKey].
```
sigurdm, line 109:
```suggestion
  /// Generate a random [AesGcmSecretKey].
```

lib/src/webcrypto/webcrypto.aesgcm.dart
```diff
@@ -14,40 +14,263 @@
...
+  /// [2]: https://en.wikipedia.org/wiki/Authenticated_encryption
   Future<Uint8List> encryptBytes(
     List<int> data,
     List<int> iv, {
     List<int>? additionalData,
     int? tagLength = 128,
   });
 
+  // TODO: Document this method, notice that [data] must be concatenation of
+  //       ciphertext and authentication tag.
```
sigurdm, line 223:
```suggestion
  // TODO: Document this method, notice that [data] must be concatenation of
  //       ciphertext and authentication tag.
  // TODO: Document what happens if the authenticity validation fails? Some Exception? 
```

sigurdm (approved):

jonasfj:
> Does this documentation come from somewhere, or you wrote it all from scratch?

I deciphered most of it from [NIST SP 800-38D][1], which is also why I leave a lot of references to it.

[1]: https://csrc.nist.gov/pubs/sp/800/38/d/final

lib/src/webcrypto/webcrypto.aesgcm.dart
jonasfj: Technically correct, but BoringSSL doesn't offer a streaming interface, I suspect I read in a bug somewhere that it doesn't because the authors thought it would be a bad idea.

So that is the opinion we're carrying into `package:webcrypto`.

Also Web Cryptography specification doesn't offer streaming APIs at all, so I only added them because BoringSSL does. And on web we always buffer up the entire stream before we do the operation :rofl:

lib/src/webcrypto/webcrypto.aesgcm.dart
jonasfj: That's a really good TODO, kind of inline with https://github.com/google/webcrypto.dart/issues/55

lib/src/webcrypto/webcrypto.aesgcm.dart
jonasfj: Yes, maybe it's messy and we should add a main, or omit imports.

Feel free to file an issue on the repository.

lib/src/webcrypto/webcrypto.aesgcm.dart
sigurdm: I'd either remove the explanation or link to the discussion. I don't think we need to explain this.

I kind-of agree that a streaming interface would be hard to use correctly. But so is basically all crypto primitives. You really have to know what you are doing.

lib/src/webcrypto/webcrypto.aesgcm.dart
sigurdm: I think this is slightly different. Because in #55 you can programmatically avoid the error (in Dart terms this is an `Error`)

However when you decrypt, there is no way of knowing up-front if the ciphertext is valid (that's the whole point of the validation) (in Dart terms this is an `Exception`) and as such documenting it seems extra important.

You could also consider returning `null` in that case?

jonasfj:
Filed https://github.com/google/webcrypto.dart/issues/102 to follow up with more recommendations, once we find credible sources we can reference.

Issue #101, opened 2024-05-02
Make dartdoc examples self-contained.


sigurdm:
The code in dartdoc examples mixes imports and statement-level code.

Ideally it would contain self-contained code that one could copy and run for one-self. And that we could extract programmatically and format, analyze etc.

PR #103, opened 2024-05-05
docs: add example section for HmacSecretKey class

HamdaanAliQuatil:
- [x] Added Examples section for the HmacSecretKey class in the API Reference Docs
- [x] Made the sample code self-contained which one could copy and run.

Here's the self-contained code and the terminal logs: https://gist.github.com/HamdaanAliQuatil/bd32945af3091ff7ba72071386ff4de7

### DIFF
--- a/lib/src/webcrypto/webcrypto.hmac.dart
+++ b/lib/src/webcrypto/webcrypto.hmac.dart
@@ -25,7 +25,44 @@ part of 'webcrypto.dart';
 ///  * [JWK] format using [HmacSecretKey.importJsonWebKey].
 ///
 /// A random key can also be generated using [HmacSecretKey.generateKey].
-///
+/// 
+/// **Example**
+/// ```dart
+/// import 'package:webcrypto/webcrypto.dart';
+/// import 'dart:convert';
+/// 
+/// Future<void> main() async {
+///   // Generate a new key using SHA-256 and an optional length parameter.
+///   final key = await HmacSecretKey.generateKey(Hash.sha256, length: 256);
+///   
+///   // Sign the message.
+///   final signature = await key.signBytes(utf8.encode('Hello World!'));
+/// 
+///   // Verify the signature.
+///   final verified = await key.verifyBytes(signature, utf8.encode('Hello World!'));
+///   assert(verified == true, 'Signature should be valid');
+/// 
+///   // Sign a stream of data.
+///   final signatureStream = await key.signStream(Stream.fromIterable([
+///     utf8.encode('Hello '),
+///     utf8.encode('World!'),
+///   ]));
+/// 
+///   // Verify the signature.
+///   final verifiedStream = await key.verifyStream(signatureStream, Stream.fromIterable([
+///     utf8.encode('Hello '),
+///     utf8.encode('World!'),
+///   ]));
+///   assert(verifiedStream == true, 'Signature should be valid');
+/// 
+///   // Export the key as raw bytes.
+///   final rawKey = await key.exportRawKey();
+/// 
+///  // Export the key as a JSON Web Key.
+///  final jwk = await key.exportJsonWebKey();
+/// }
+/// ```
+/// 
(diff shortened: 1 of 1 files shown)


HamdaanAliQuatil:
Takes a step towards #101

lib/src/webcrypto/webcrypto.hmac.dart
```diff
@@ -25,7 +25,44 @@ part of 'webcrypto.dart';
...
 /// A random key can also be generated using [HmacSecretKey.generateKey].
-///
+/// 
+/// **Example**
+/// ```dart
+/// import 'package:webcrypto/webcrypto.dart';
+/// import 'dart:convert';
+/// 
+/// Future<void> main() async {
+///   // Generate a new key using SHA-256 and an optional length parameter.
```
jonasfj, line 35: I can't parse this comment.

The length parameter is optional, but that's irrelevant, the comment should probably explain what the value passed is.

Also omitting length is fine, in this case it has no effect.

I suggest that we leave length of the example.

lib/src/webcrypto/webcrypto.hmac.dart
```diff
@@ -25,7 +25,44 @@ part of 'webcrypto.dart';
...
+///   final key = await HmacSecretKey.generateKey(Hash.sha256, length: 256);
+///   
+///   // Sign the message.
+///   final signature = await key.signBytes(utf8.encode('Hello World!'));
+/// 
+///   // Verify the signature.
+///   final verified = await key.verifyBytes(signature, utf8.encode('Hello World!'));
+///   assert(verified == true, 'Signature should be valid');
+/// 
+///   // Sign a stream of data.
```
jonasfj, line 45: Let's omit the streaming API...

We don't want to give the impression that its different...

Otherwise, we could do signBytes and verifyStream if we want to be fun.

Or we could explain they produce the same result.

lib/src/webcrypto/webcrypto.hmac.dart
```diff
@@ -25,7 +25,44 @@ part of 'webcrypto.dart';
...
+///   ]));
+/// 
+///   // Verify the signature.
+///   final verifiedStream = await key.verifyStream(signatureStream, Stream.fromIterable([
+///     utf8.encode('Hello '),
+///     utf8.encode('World!'),
+///   ]));
+///   assert(verifiedStream == true, 'Signature should be valid');
+/// 
+///   // Export the key as raw bytes.
```
jonasfj, line 58: Let's omit exporting, again let's keep example small..

Or if we want use one of the export methods not both.

jonasfj:
The comment about length is a little confusing. The rest looks fine, my comments are just nits :)

lib/src/webcrypto/webcrypto.hmac.dart
HamdaanAliQuatil: Removed the length parameter.
Improved the comment

lib/src/webcrypto/webcrypto.hmac.dart
HamdaanAliQuatil: Removed signStream and verifyStream

lib/src/webcrypto/webcrypto.hmac.dart
HamdaanAliQuatil: Removed exportRawKey

jonasfj (approved):

lib/src/webcrypto/webcrypto.hmac.dart
jonasfj: It's possible that in all classes that have `xxxStream` and `xxxBytes` we should make a note that they give the same result, given the same list of bytes :D

Though, I'm not sure it's necessary, it's kind of obvious, and we probably point it out a few places too.

But it might be nice to make tutorial style documentation thing about the streaming API, perhaps.

PR #110, opened 2024-05-07
Override toString for All Private Classes 

### LINKED ISSUE #104: Add .toString() methods on all private classes
It's sad when `print(HmacSecretKey.generateKey(Hash.sha256))` shows `Instance of '_HmacSecretKey'`.

Perhaps we should override the `toString()` method on all the private classes.

The user doesn't need to know about `_HmacSecretKey`, since it's a private class.
And we could probably display something more useful for debugging purposes.

I haven't figured this out yet, but we could do something like:
 * `HmacSecretKey(hash: Hash.sha256, key: ***)`
 * `HmacSecretKey(Hash.sha256)`
 * `Instance of 'HmacSecretKey with Hash.sha256'`
 * `Instance of 'HmacSecretKey'`
 * `HmacSecretKey with sha256`
 * or ???

It could be useful if print was helpful when debugging stuff :D

On the flip side, the `HmacSecretKey` doesn't have a `hash` property, so if you have an instance of `HmacSecretKey` and you want to know what hash it uses, the only thing you can do is:
 * (a) Search your code base, or,
 * (b) Use `HmacSecretKey.toString().contains('sha256')`, if we added a fancy variant of `toString()`.

I imagine that we'd prefer to avoid using doing (b). Not necessarily be adding a `hash` property they don't need, but by forcing them to do (a) :rofl: 

I think the point I'm trying to make is that, if we add to much information in `toString()`, then maybe people will use to inspect an object at runtime and rely on the behavior of `toString()`.
And I don't think we want people to rely on the behavior of `toString()`.

So maybe it's safest to just do:
`macSecretKey.generateKey(Hash.sha256).toString() == "Instance of 'HmacSecretKey'"`

It's boring, it's simple, but it doesn't leak any information that could lure a user into misusing `toString()`.
I could ofcourse be convinced otherwise.

HamdaanAliQuatil:
Closes #104 

Updated the following Private Functions:

- [x] _HmacSecretKey
- [x] _AesCbcSecretKey
- [x] _AesCtrSecretKey
- [x] _AesGcmSecretKey
- [x] _EcdhPublicKey/ PrivateKey
- [x] _EcdsaPublicKey/ PrivateKey
- [x] _HkdfSecretKey
- [x] _Pbkdf2SecretKey
- [x] _RsaOaepPublicKey
- [x] _RsaPssPublicKey
- [x] _RsassaPkcs1V15PublicKey

No issues found in:
- _Hash

### DIFF
--- a/lib/src/impl_ffi/impl_ffi.aescbc.dart
+++ b/lib/src/impl_ffi/impl_ffi.aescbc.dart
@@ -98,6 +98,11 @@ class _AesCbcSecretKey implements AesCbcSecretKey {
   _AesCbcSecretKey(this._key);
 
   @override
+  String toString() {
+    return 'Instance of \'AesCbcSecretKey\'';
+  }
+
+  @override
   Future<Uint8List> decryptBytes(List<int> data, List<int> iv) async =>
       await _bufferStream(decryptStream(Stream.value(data), iv));
 
--- a/lib/src/impl_ffi/impl_ffi.aesctr.dart
+++ b/lib/src/impl_ffi/impl_ffi.aesctr.dart
@@ -204,6 +204,11 @@ class _AesCtrSecretKey implements AesCtrSecretKey {
   final Uint8List _key;
   _AesCtrSecretKey(this._key);
 
+  @override
+  String toString() {
+    return 'Instance of \'AesCtrSecretKey\'';
+  }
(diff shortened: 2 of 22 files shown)


jonasfj (approved):
r+

PR #114, opened 2024-05-13
Document all members of EcdhPrivateKey Class

HamdaanAliQuatil:


lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,223 @@
...
+/// two parties to establish a shared secret over an insecure channel.
+/// 
+/// An [EcdhPrivateKey] instance represents a private key that can be used to
+/// derive a shared secret with a public key.
+/// 
+/// Instances of [EcdhPrivateKey] can be imported from:
+/// * PKCS8 Key using [EcdhPrivateKey.importPkcs8Key], and,
+/// * JSON Web Key using [EcdhPrivateKey.importJsonWebKey].
+/// 
+/// A random key pair can also be generated using [EcdhPrivateKey.generateKey].
```
jonasfj, line 27:
```suggestion
/// A random key pair can be generated using [EcdhPrivateKey.generateKey].
```
But I might also be careful to say random, I think that's actually what the `AesCbcSecretKey` is. But it's a verb I would be careful of using around a key pair. Since the private key does match the public key, is it random?

I'd suggest something like we have on [RsaPssPrivateKey](https://pub.dev/documentation/webcrypto/latest/webcrypto/RsaPssPrivateKey-class.html) or [RsassaPkcs1V15PrivateKey](https://pub.dev/documentation/webcrypto/latest/webcrypto/RsassaPkcs1V15PrivateKey-class.html).

```suggestion
/// A public-private [KeyPair] consisting of a [EcdhPublicKey] and
/// a [EcdhPrivateKey] can be generated using [EcdhPrivateKey.generateKey].
```

I don't know if the private key in ECDH can be called random. That might actually be possible. I think in RSA, it would be better to not say random, because an RSA key is derived from random input. But the actual key consists of primes, so a key is not random data.

In AES I think we can say random, because the key generation logic is just generating random bytes :rofl: 

Unless, we're sure the same applies here, I would avoid making the claim.

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,223 @@
...
+/// **Example**
+/// ```dart
+/// import 'package:webcrypto/webcrypto.dart';
+/// 
+/// Future<void> main() async {
+///   // Generate a new key pair using the P-256 curve.
+///   final keyPair = await EcdhPrivateKey.generateKey(EllipticCurve.p256);
+///   
+///   // Derive 256 bits from the private key using the public key.
+///   final derivedBits = await keyPair.privateKey.deriveBits(256, keyPair.publicKey);
```
jonasfj, line 38: This isn't what ECDH is used for right?

```dart
// Alice generates a key-pair
final kpA = await EcdhPrivateKey.generateKey(EllipticCurve.p256);

// Bob generates a key-pair
final kpB = await EcdhPrivateKey.generateKey(EllipticCurve.p256);

// If Alice and Bob exchanges public keys, they can generate the same shared secret
final sharedSecretA = await pkA.privateKey.deriveBits(256, pkB.publicKey);
final sharedSecretB = await pkB.privateKey.deriveBits(256, pkA.publicKey);
assert(base64.encode(sharedSecretA) == base64.encode(sharedSecretB));
```

Or am I wrong here? maybe we should do a test to make sure we understand how this is supposed to work.

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,223 @@
...
+///   // Export the private key as a JSON Web Key.
+///   final exportedPrivateKey = await keyPair.privateKey.exportJsonWebKey();
+/// }
+/// ```
 @sealed
 abstract class EcdhPrivateKey {
   EcdhPrivateKey._(); // keep the constructor private.
 
-  // Note. unsupported on Firefox, see EcdsaPrivateKey.importPkcs8Key
+  /// Import [EcdhPrivateKey] using the PKCS8 format[1].
```
jonasfj, line 48: You can't say that you're using a format.

Something is a format.

See: https://pub.dev/documentation/webcrypto/latest/webcrypto/RsaPssPrivateKey/importPkcs8Key.html

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,223 @@
...
+  ///   );
+  /// }
+  /// ```
+  /// 
+  /// [1]: https://datatracker.ietf.org/doc/html/rfc5208
   static Future<EcdhPrivateKey> importPkcs8Key(
     List<int> keyData,
     EllipticCurve curve,
   ) {
+  // Note. unsupported on Firefox, see EcdsaPrivateKey.importPkcs8Key
```
jonasfj, line 74: This should be noted in a remark similar to what we have at the bottom of:
https://pub.dev/documentation/webcrypto/latest/webcrypto/AesCtrSecretKey/encryptBytes.html

I think the "see EcdsaPrivateKey.importPkcs8Key" is just a note to us, because we probably wrote more about it there. Hopefully, the source contains a bug we can reference.

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,223 @@
...
     return impl.ecdhPrivateKey_importPkcs8Key(keyData, curve);
   }
 
+  /// Import [EcdhPrivateKey] from a [JSON Web Key][1].
```
jonasfj, line 78: Check:
https://pub.dev/documentation/webcrypto/latest/webcrypto/RsassaPkcs1V15PrivateKey/importJsonWebKey.html

and

https://pub.dev/documentation/webcrypto/latest/webcrypto/RsaPssPrivateKey/importJsonWebKey.html

Though this is pretty good :rofl:

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,223 @@
...
+  /// parameter.
+  /// * `"d"`: The private key for the Elliptic Curve point represented as a
+  /// base64url encoded string. The length of this octet string MUST be
+  /// 32 bytes for `"crv": "P-256"`, 48 bytes for `"crv": "P-384"`, and
+  /// 66 bytes for `"crv": "P-521"`.
+  /// 
+  /// For importing a JWK with:
+  /// * `"crv": "P-256"`, use [EllipticCurve.p256]
+  /// * `"crv": "P-384"`, use [EllipticCurve.p384]
+  /// * `"crv": "P-521"`, use [EllipticCurve.p521]
```
jonasfj, line 100:
```suggestion
  /// For importing a JWK with:
  /// * `"crv": "P-256"`, use [EllipticCurve.p256],
  /// * `"crv": "P-384"`, use [EllipticCurve.p384], and,
  /// * `"crv": "P-521"`, use [EllipticCurve.p521].
```

lib/src/webcrypto/webcrypto.ecdh.dart
jonasfj, line 78:
```suggestion
  /// Import ECDH private key in [JSON Web Key][1] format.
```

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,223 @@
...
   static Future<EcdhPrivateKey> importPkcs8Key(
     List<int> keyData,
     EllipticCurve curve,
   ) {
+  // Note. unsupported on Firefox, see EcdsaPrivateKey.importPkcs8Key
     return impl.ecdhPrivateKey_importPkcs8Key(keyData, curve);
   }
 
+  /// Import [EcdhPrivateKey] from a [JSON Web Key][1].
+  ///
```
jonasfj, line 79: We probably want something like:
> The jwk should be given as [Map](https://api.flutter.dev/flutter/dart-core/Map-class.html), [String](https://api.flutter.dev/flutter/dart-core/String-class.html), [List](https://api.flutter.dev/flutter/dart-core/List-class.html) the same way [jsonDecode](https://api.flutter.dev/flutter/dart-convert/jsonDecode.html) from dart:convert represents decoded JSON values.

See: https://pub.dev/documentation/webcrypto/latest/webcrypto/RsaPssPrivateKey/importJsonWebKey.html

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,223 @@
...
   }
 
+  /// Import [EcdhPrivateKey] from a [JSON Web Key][1].
+  /// 
+  /// JSON Web Keys imported using [EcdhPrivateKey.importJsonWebKey] must 
+  /// have `"kty": "EC"` and MUST have the following parameters:
+  /// * `"crv"`: The curve used with the key. This MUST match the curve
+  ///  parameter.
+  /// * `"x"`: The x coordinate for the Elliptic Curve point represented
+  /// as a base64url encoded string. The length of this octet string MUST
```
jonasfj, line 85:
```suggestion
  /// as a [base64Url] encoded string. The length of this octet string MUST
```

We can link to: https://api.dart.dev/stable/3.3.4/dart-convert/base64Url-constant.html

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,223 @@
...
+  /// be the full size of a coordinate for the curve specified in the `"crv"`
+  /// parameter.
+  /// * `"y"`: The y coordinate for the Elliptic Curve point represented
+  /// as a base64url encoded string. The length of this octet string MUST
+  /// be the full size of a coordinate for the curve specified in the `"crv"`
+  /// parameter.
+  /// * `"d"`: The private key for the Elliptic Curve point represented as a
+  /// base64url encoded string. The length of this octet string MUST be
+  /// 32 bytes for `"crv": "P-256"`, 48 bytes for `"crv": "P-384"`, and
+  /// 66 bytes for `"crv": "P-521"`.
```
jonasfj, line 95: Is `d` the private key, or part of it.

We could also consider being less specific and simply stating the JWK must have `x`, `y` and `d` as based64 URL encoded strings.

jonasfj (commented):
Mostly just small things. I'll probably have to spend a bit more time on the review.

I like that we have documentation which explains a lot of things, we just have to careful we don't make mistakes. And I'll freely admit that I have to lookup things before I'm sure what is correct and what is not.

lib/src/webcrypto/webcrypto.ecdh.dart
HamdaanAliQuatil: Makes much more sense. I was only providing an example of how the function is used instead of how ECDH should be used. I'll update the example with this

lib/src/webcrypto/webcrypto.ecdh.dart
HamdaanAliQuatil: Alright. Ig a correct way to put it would be:

Import `[EcdhPrivateKey]` in the PKCS8 format[1].

Does this look good?

lib/src/webcrypto/webcrypto.ecdh.dart
HamdaanAliQuatil: Moved the Note above the function definition. Removed "see EcdsaPrivateKey.importPkcs8Key"

lib/src/webcrypto/webcrypto.ecdh.dart
HamdaanAliQuatil: Improved in the second Iteration. Resolving as outdated

Updated Review: https://github.com/google/webcrypto.dart/pull/114#discussion_r1598096846

lib/src/webcrypto/webcrypto.ecdh.dart
HamdaanAliQuatil: Added the {@macro importJsonWebKey:jwk} 
TODO: Remove `[hash]` out of the template in next PR

lib/src/webcrypto/webcrypto.ecdh.dart
HamdaanAliQuatil: Leaving `d` as the private key after our discussion.

lib/src/webcrypto/webcrypto.ecdh.dart
jonasfj: What I meant was that we should probably note in the documentation comment that it won't work for Firefox.

If we actually lookup the code for what the comment says, we'll see:
https://github.com/google/webcrypto.dart/blob/master/lib/src/webcrypto/webcrypto.ecdsa.dart#L21-L29

That comment sends us to:
https://bugzilla.mozilla.org/show_bug.cgi?id=1133698

Which appears to be fixed 3 years ago :tada:, so maybe `importPkcs8Key` works for EC keys on Firefox. We should update our tests to see if this is the case.

If `importPkcs8Key` works, then we won't need to document that it doesn't work on Firefox.

Filed: https://github.com/google/webcrypto.dart/pull/117 to remove this comment.
As it now appears to work.

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,231 @@
...
     List<int> keyData,
     EllipticCurve curve,
   ) {
+  // Note. unsupported on Firefox, see EcdsaPrivateKey.importPkcs8Key
     return impl.ecdhPrivateKey_importPkcs8Key(keyData, curve);
   }
 
+  /// Import ECDH private key in [JSON Web Key][1] format.
+  /// {@macro importJsonWebKey:jwk}
+  ///
```
jonasfj, line 87:
```suggestion
  /// Import ECDH private key in [JSON Web Key][1] format.
  ///
  /// {@macro importJsonWebKey:jwk}
  /// 
```

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,231 @@
...
+  /// [1]: https://www.rfc-editor.org/rfc/rfc7518.html#section-6.2
   static Future<EcdhPrivateKey> importJsonWebKey(
     Map<String, dynamic> jwk,
     EllipticCurve curve,
   ) {
     return impl.ecdhPrivateKey_importJsonWebKey(jwk, curve);
   }
 
+  /// Generate a new [EcdhPrivateKey] and [EcdhPublicKey] pair.
+  /// The [curve] parameter specifies the curve to use for the key pair.
```
jonasfj, line 145:
```suggestion
  /// Generate a new [EcdhPrivateKey] and [EcdhPublicKey] pair.
  ///
  /// The [curve] parameter specifies the curve to use for the key pair.
```

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,231 @@
...
+  /// 
+  /// To generate a key pair using the P-256 curve, use [EllipticCurve.p256].
+  /// To generate a key pair using the P-384 curve, use [EllipticCurve.p384].
+  /// To generate a key pair using the P-521 curve, use [EllipticCurve.p521].
```
jonasfj, line 149:
```dart
  ///  * [EllipticCurve.p256] for P-256 curve.
  ...
```

Or something else, but definitely more concise.

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,231 @@
...
+  /// ```
+  /// 
   static Future<KeyPair<EcdhPrivateKey, EcdhPublicKey>> generateKey(
     EllipticCurve curve,
   ) {
     return impl.ecdhPrivateKey_generateKey(curve);
   }
 
+  /// Derive an array of bits from the [EcdhPrivateKey].
+  /// The [length] parameter specifies the number of bits to derive and
```
jonasfj, line 176:
```suggestion
  /// Derive an array of bits from the [EcdhPrivateKey].
  ///
  /// The [length] parameter specifies the number of bits to derive and 
```

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,231 @@
...
   // length from the curve. Note p512 can provide up to: 528 bits!!!
   //
   // See: https://tools.ietf.org/html/rfc6090#section-4
   // Notice that this is not uniformly distributed, see also:
   // https://tools.ietf.org/html/rfc6090#appendix-B
-  Future<Uint8List> deriveBits(int length, EcdhPublicKey publicKey);
 
-  // Note. unsupported on Firefox, see EcdsaPrivateKey.importPkcs8Key
+  /// Export the [EcdhPrivateKey] as a PKCS8 key.
+  /// This returns the private key as an octet string.
```
jonasfj, line 203:
```suggestion
  /// Export the [EcdhPrivateKey] as a PKCS8 key.
  ///
  /// This returns the private key as an octet string.
```

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,231 @@
...
+  /// 
+  // Note. unsupported on Firefox
   Future<Uint8List> exportPkcs8Key();
 
+  /// Export the [EcdhPrivateKey] and [EcPublicKey] as a [JSON Web Key][1].
+  /// 
+  /// **Example**
+  /// ```dart
+  /// import 'package:webcrypto/webcrypto.dart';
+  ///
```
jonasfj, line 227: Check documentation for the other `exportJsonWebKey` methods. There is probably a macro describing what comes out.

jonasfj (changes requested):

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,231 @@
...
+  /// The [keyData] parameter is an octet string whose 
+  /// contents are the value of the private key.
+  /// The [curve] parameter specifies the curve to use 
+  /// for the key pair.
+  /// 
+  /// **Example**
+  /// ```dart
+  /// import 'package:webcrypto/webcrypto.dart';
+  /// 
+  /// final pkcs8Key = [ /* ... */ ];
```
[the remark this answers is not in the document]
HamdaanAliQuatil: 1. Generated Private key using `ssh-keygen -t ecdsa -b 256 -m PKCS8`
2. Used the Private key in `PemCodec` function. Tested the running code.
3. Updated the Example

lib/src/webcrypto/webcrypto.ecdh.dart
HamdaanAliQuatil: Added ```{@macro exportJsonWebKey:returns}```

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,231 @@
...
 @sealed
 abstract class EcdhPrivateKey {
   EcdhPrivateKey._(); // keep the constructor private.
 
-  // Note. unsupported on Firefox, see EcdsaPrivateKey.importPkcs8Key
+  /// Import [EcdhPrivateKey] in the PKCS8 format[1].
+  /// The [keyData] parameter is an octet string whose 
+  /// contents are the value of the private key.
+  /// The [curve] parameter specifies the curve to use 
+  /// for the key pair.
```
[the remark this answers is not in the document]
HamdaanAliQuatil: Noted for upcoming work 🫡

lib/src/webcrypto/webcrypto.ecdh.dart
jonasfj: Perfect!

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,250 @@
...
   //
   // See: https://tools.ietf.org/html/rfc6090#section-4
   // Notice that this is not uniformly distributed, see also:
   // https://tools.ietf.org/html/rfc6090#appendix-B
-  Future<Uint8List> deriveBits(int length, EcdhPublicKey publicKey);
 
-  // Note. unsupported on Firefox, see EcdsaPrivateKey.importPkcs8Key
+  /// Export the [EcdhPrivateKey] as a PKCS8 key.
+  ///
+  /// This returns the private key as an octet string.
```
jonasfj, line 220: I would just say: returns private key as a list of bytes.

Unless, we're really fancy. Look at `RsassaPkcs1V15PrivateKey.exportPkcs8Key` it says:
> Returns the DER encoding of the PrivateKeyInfo structure specified in [RFC 5208](https://tools.ietf.org/html/rfc5208) as a list of bytes.

I assume I really did look up the specification to find out exactly what is returned and in what format.

I'm guessing it might be a different thing for EC keys.

--------------

I actually think I did this, because normally when someone says "PKCS #8" format, you're thinking about a base64 encoded string inside the PEM `---- PRIVATE KEY---` ASCII armor thing..

Which is not what this returns.

Look at: https://www.w3.org/TR/WebCryptoAPI/#ecdh-operations

So it might be worth saying what structure it returns in what encoding as bytes.

I suspect this is also where we have to at-least for our own sanity read about BER vs DER (I forgot all about this).
See: https://en.wikipedia.org/wiki/X.690

But I think webcrypto might only support DER encoding (this might be something we should test across platforms, probably something we should file an issue for testing :rofl: )

--------------------

So, yeah, I'm not exactly sure what this should say. We should probably read a bit more in the RFCs and webcrypto spec to make sure what references we want to give.

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,250 @@
...
   static Future<KeyPair<EcdhPrivateKey, EcdhPublicKey>> generateKey(
     EllipticCurve curve,
   ) {
     return impl.ecdhPrivateKey_generateKey(curve);
   }
 
+  /// Derive an array of bits from the [EcdhPrivateKey].
+  ///
+  /// The [length] parameter specifies the number of bits to derive and 
+  /// should be multiples of 8.
```
jonasfj, line 193: What happens if it's not a multiple of 8? Is it an error? What does the webcrypto specification say?

Must it be a multiple of 8, then we should say must :D
```suggestion
  /// The [length] parameter specifies the number of bits to derive and 
  /// should be multiples of 8.
```

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,250 @@
...
+  /// **Example**
+  /// ```dart
+  /// import 'package:webcrypto/webcrypto.dart';
+  /// 
+  /// Future<void> main() async {
+  ///   // Generate a new key pair using the P-256 curve.
+  ///   final keyPair = await EcdhPrivateKey.generateKey(EllipticCurve.p256);
+  /// 
+  ///   // Derive 256 bits from the private key using the public key.
+  ///   final derivedBits = await keyPair.privateKey.deriveBits(256, keyPair.publicKey);
```
jonasfj, line 205: When is usage such as this legitimate?

Is it wrong to use the same public / private key pair together?
I don't know, but I thought the idea was that you generated two key-pairs, exchanged public keys and then generated the same shared secret.

Using a single key-pair may work, but what is the point? If you're sending the private key over a secure channel, then you have a secure channel and might as well send a randomly generated shared secret instead.

At best this is an example of a fairly complex and expensive RNG, at worst this is a flared RNG.
I don't have to expertise to say whether this is a reasonable way to generate random numbers. But I don't think this is how ECDH is typically intended to be used.

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,250 @@
...
+  /// }
+  /// ```
+  /// 
   static Future<KeyPair<EcdhPrivateKey, EcdhPublicKey>> generateKey(
     EllipticCurve curve,
   ) {
     return impl.ecdhPrivateKey_generateKey(curve);
   }
 
+  /// Derive an array of bits from the [EcdhPrivateKey].
```
jonasfj, line 190: "an array of bits" is just type information. It's more accurate to write `UInt8List` which is the return type. So repeating doesn't really help much.

Look at:
https://pub.dev/documentation/webcrypto/latest/webcrypto/HkdfSecretKey/deriveBits.html

We could say `Derive shared secret from ...` or `Derive key from ...`.

Also we don't derive from private key, you derive from private and public key.
Or derive from two key-pairs might be more accurate, right?

You have the private key from one pair and the private key from another, hence, you derive from two key-pairs.

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,250 @@
...
+  ///
+  /// The [length] parameter specifies the number of bits to derive and 
+  /// should be multiples of 8.
+  /// The [publicKey] parameter is the [EcdhPublicKey].
```
jonasfj, line 194: This just states the type.

I do think we need to clearly state what the algorithm does.

Maybe, we should concisely explain that it derives a shared secret from two key-pairs.
Using the private key from one and public key from the other. And that using the private key from the other and public key from the one, will result in the same shared secret.

Maybe, that could get confusing, so maybe it's sensible to say: "Derives a shared-secret from two ECDH key-pairs A and B, using the private key from A and public key from B. Using the public key from A and the private key from B will result in the same shared secret. Allowing two parties to obtain a shared secret by exchanging public keys.

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -14,43 +14,250 @@
...
   // null for length (in this primitive). However, you can always know the right
   // length from the curve. Note p512 can provide up to: 528 bits!!!
   //
   // See: https://tools.ietf.org/html/rfc6090#section-4
   // Notice that this is not uniformly distributed, see also:
   // https://tools.ietf.org/html/rfc6090#appendix-B
-  Future<Uint8List> deriveBits(int length, EcdhPublicKey publicKey);
 
-  // Note. unsupported on Firefox, see EcdsaPrivateKey.importPkcs8Key
+  /// Export the [EcdhPrivateKey] as a PKCS8 key.
```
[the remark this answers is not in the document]
HamdaanAliQuatil: Updated

HamdaanAliQuatil:
Splitting this into smaller PRs

HamdaanAliQuatil:
Closing as all Docs are split into smaller PRs

PR #119, opened 2024-05-22
Attempt to reduce iOS integration test timeouts

jonasfj:
Trying out the suggestion from:
https://github.com/flutter/flutter/issues/105913#issuecomment-1500767101

Probably I should gather more data, to know if it's often remotely close to taking 12 minutes (which would be surprising).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -110,6 +110,10 @@ jobs:
         with:
           model: 'iPhone 15'
       - run: flutter pub get
+      - run: flutter pub get
+        working-directory: ./example
+      - run: flutter build ios --simulator --target=integration_test/webcrypto_test.dart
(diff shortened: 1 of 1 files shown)


jonasfj:
Looking a previous runs, usually it takes 4 minutes to run the iPhone integration tests on a simulator.

So if it times out at 12 minutes, it's probably some intermittent bug.
Could be that pre-building reduces the risk of that bug happening, perhaps we're eating up too much memory or something.

jonasfj:
Worked once, trying again :D

jonasfj:
It appears to take a long time on second attempt, I'm guessing it'll timeout.
But what is interesting, is that it looks like it's build step that hangs:
image

Not the step that runs integration tests.

PR #120, opened 2024-05-23
Add support for AGP 8.4

emersion:
Fixes the following errors with AGP 8.4:

    A problem occurred configuring project ':webcrypto'.
    > [CXX1110] Platform version 16 is unsupported by this NDK. Please change minSdk to at least 21 to avoid undefined behavior. To suppress this error, add android.ndk.suppressMinSdkVersionError=21 to the project's gradle.properties or set android.experimentalProperties["android.ndk.suppressMinSdkVersionError"]=21 in the Gradle build file.

and

    Execution failed for task ':webcrypto:processReleaseManifest'.
    > A failure occurred while executing com.android.build.gradle.tasks.ProcessLibraryManifest$ProcessLibWorkAction
       > Incorrect package="com.example.webcrypto" found in source AndroidManifest.xml: /home/simon/src/webcrypto.dart/android/src/main/AndroidManifest.xml.
         Setting the namespace via the package attribute in the source AndroidManifest.xml is no longer supported.
         Recommendation: remove package="com.example.webcrypto" from the source AndroidManifest.xml: /home/simon/src/webcrypto.dart/android/src/main/AndroidManifest.xml.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -25,10 +25,7 @@ apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
 android {
-    // Conditional for compatibility with AGP <4.2.
-    if (project.android.hasProperty("namespace")) {
-        namespace 'dev.google.webcrypto'
-    }
+    namespace 'dev.google.webcrypto'
 
     compileSdkVersion 31
 
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -53,7 +53,7 @@ android {
     }
 
     defaultConfig {
-        minSdkVersion 16
+        minSdkVersion 21
         ndk {
             abiFilters 'armeabi-v7a', 'arm64-v8a', 'x86', 'x86_64'
         }
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -121,6 +121,10 @@ jobs:
         with:
           channel: 'stable'
           cache: true
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'zulu'
+          java-version: 21
       - name: Configure Flutter
         run: |
           flutter config --no-analytics
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -14,6 +14,5 @@
  limitations under the License.
 -->
 
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-  package="com.example.webcrypto">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
 </manifest>
--- a/example/android/app/build.gradle
+++ b/example/android/app/build.gradle
@@ -58,6 +58,7 @@ android {
             signingConfig signingConfigs.debug
         }
     }
+    namespace 'com.example.webcrypto_example'
 }
 
 flutter {
--- a/example/android/app/build.gradle
+++ b/example/android/app/build.gradle
@@ -45,7 +45,7 @@ android {
     defaultConfig {
         // TODO: Specify your own unique Application ID (https://developer.android.com/studio/build/application-id.html).
         applicationId "com.example.webcrypto_example"
-        minSdkVersion 16
+        minSdkVersion flutter.minSdkVersion
         targetSdkVersion 31
         versionCode flutterVersionCode.toInteger()
         versionName flutterVersionName
--- a/example/android/app/src/debug/AndroidManifest.xml
+++ b/example/android/app/src/debug/AndroidManifest.xml
@@ -14,8 +14,7 @@
  limitations under the License.
 -->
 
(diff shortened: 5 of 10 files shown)


android/build.gradle
```diff
@@ -53,7 +50,7 @@ android {
     }
 
     defaultConfig {
-        minSdkVersion 16
+        minSdkVersion 21
```
emersion, line 53: I wonder if we should just use `flutter.minSdkVersion` somehow? Did not seem to work in my testing though:

```
> Could not get unknown property 'flutter' for extension 'android' of type com.android.build.gradle.LibraryExtension.
```

android/build.gradle
```diff
@@ -25,10 +25,7 @@ apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
 android {
-    // Conditional for compatibility with AGP <4.2.
-    if (project.android.hasProperty("namespace")) {
-        namespace 'dev.google.webcrypto'
-    }
+    namespace 'dev.google.webcrypto'
 
     compileSdkVersion 31
```
emersion, line 30: Is there a reason for this being 31 instead of `flutter.ndkVersion`? (Currently 23.1.7779620)

jonasfj:
Do I understand it correctly that this will drop support for Android 4.x, which is more than 12 years old at this point.

That might be acceptable, I'm guessing :D

android/build.gradle
```diff
@@ -25,10 +25,7 @@ apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
 android {
-    // Conditional for compatibility with AGP <4.2.
-    if (project.android.hasProperty("namespace")) {
-        namespace 'dev.google.webcrypto'
-    }
+    namespace 'dev.google.webcrypto'
```
jonasfj, line 28: Dumb question, but do we not need this anymore?

Will `minSdkVersion 21` make AGP < 4.2 irrelevant?

jonasfj:
Hmm, I'm guessing we should also look at:
https://docs.flutter.dev/release/breaking-changes/flutter-gradle-plugin-apply

I'll file it separately, as https://github.com/google/webcrypto.dart/issues/124

jonasfj (approved):
Is there some kind of testing we should do to be more confident this is correct?

I'd love to have it covered in CI.

HamdaanAliQuatil:
> Fixes the following errors with AGP 8.4:
> 
> ```
> A problem occurred configuring project ':webcrypto'.
> > [CXX1110] Platform version 16 is unsupported by this NDK. Please change minSdk to at least 21 to avoid undefined behavior. To suppress this error, add android.ndk.suppressMinSdkVersionError=21 to the project's gradle.properties or set android.experimentalProperties["android.ndk.suppressMinSdkVersionError"]=21 in the Gradle build file.
> ```
> 
> and
> 
> ```
> Execution failed for task ':webcrypto:processReleaseManifest'.
> > A failure occurred while executing com.android.build.gradle.tasks.ProcessLibraryManifest$ProcessLibWorkAction
>    > Incorrect package="com.example.webcrypto" found in source AndroidManifest.xml: /home/simon/src/webcrypto.dart/android/src/main/AndroidManifest.xml.
>      Setting the namespace via the package attribute in the source AndroidManifest.xml is no longer supported.
>      Recommendation: remove package="com.example.webcrypto" from the source AndroidManifest.xml: /home/simon/src/webcrypto.dart/android/src/main/AndroidManifest.xml.
> ```

@emersion can you please let me know how to reproduce this error

android/build.gradle
emersion: Seems like plugins can't do that.

emersion:
> Do I understand it correctly that this will drop support for Android 4.x, which is more than 12 years old at this point.

Correct. Note that the latest version of Flutter [requires Android ≥ 5.x](https://docs.flutter.dev/reference/supported-platforms), so this just aligns the plugin requirements with Flutter's.

> Hmm, I'm guessing we should also look at:
https://docs.flutter.dev/release/breaking-changes/flutter-gradle-plugin-apply

Yes, that would be nice to fix as well. But it's orthogonal.

> Is there some kind of testing we should do to be more confident this is correct?
>
> I'd love to have it covered in CI.

I've pushed a commit to upgrade the example to AGP 8.4. The failures reported in the first two commits can be reproduced there.

android/build.gradle
emersion: AGP 8.4 rejects `AndroidManifest.xml` files with a `package` attribute. If we only set the namespace for AGP ≥ 4.2, then on AGP < 4.2 we won't define any namespace at all and the build will fail anyways.

I think it's reasonable to ask library users to upgrade AGP ≥ 4.2 when upgrading webcrypto.dart.

PR #121, opened 2024-05-24
Document importPkcs8Key in the EcdhPrivateKey Class

HamdaanAliQuatil:


HamdaanAliQuatil:
Split from #114

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -18,6 +18,41 @@ part of 'webcrypto.dart';
 abstract class EcdhPrivateKey {
   EcdhPrivateKey._(); // keep the constructor private.
 
+  /// Import [EcdhPrivateKey] in the [PKCS #8][1] format.
+  ///
+  /// The [keyData] parameter is an octet string whose 
+  /// contents are the value of the private key.
+  /// The [curve] parameter specifies the curve to use 
+  /// for the key pair.
```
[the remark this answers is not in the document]
jonasfj: It might be better to just say:
> The [curve] specified must match the curved used in [keyData].

Let's also add a test case, or at-least manually test that if [curve] is different from the curve the key was generated with then importing it won't work. I'm guessing we could test this on across all platforms and browsers.

We could just file an issue to test this, I think that'd be fine for now. And maybe manually try it out, so we know that it will in fact fail.

I think it'll because the curve is probably encoded in the PKCS #8 format, right?

Looking up the operation: https://www.w3.org/TR/WebCryptoAPI/#ecdh-operations that's pretty clear.
image

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -18,6 +18,41 @@ part of 'webcrypto.dart';
 abstract class EcdhPrivateKey {
   EcdhPrivateKey._(); // keep the constructor private.
 
+  /// Import [EcdhPrivateKey] in the [PKCS #8][1] format.
+  ///
+  /// The [keyData] parameter is an octet string whose 
+  /// contents are the value of the private key.
```
jonasfj, line 24: What is keyData? 

See: https://www.w3.org/TR/WebCryptoAPI/#ecdh-operations > "Import Key" > If format is "pkcs8".

Or look at the export operation too.

-------------

Btw, looks like https://www.w3.org/TR/WebCryptoAPI/#ecdh-operations
refers to "privateKeyInfo ASN.1 structure defined in RFC 5280", but RFC 5208 is the one that defines privateKeyInfo, see:

https://www.ietf.org/rfc/rfc5208.txt

-------------------

The more I read I think the input is: `privateKeyInfo` structure.
Also reading: https://www.w3.org/TR/WebCryptoAPI/#concept-parse-an-asn1-structure

I get the impression that it has to be a DER encoded `privateKeyInfo` structure.

which is similar to what we stated in:
https://pub.dev/documentation/webcrypto/latest/webcrypto/RsassaPkcs1V15PrivateKey/importPkcs8Key.html

I don't think we have to specify that all the sub-properties of the `privateKeyInfo` must be.

But we might want to say something about the curve must match what is specified in ......

lib/src/webcrypto/webcrypto.ecdh.dart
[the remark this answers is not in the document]
HamdaanAliQuatil: Updated the doc. Will add a new test in a separate PR

lib/src/webcrypto/webcrypto.ecdh.dart
HamdaanAliQuatil: Can we write something like - 
[keyData] is the DER encoding of the PrivateKeyInfo structure specified in [RFC 5208](https://tools.ietf.org/html/rfc5208). The [privateKeyAlgorithm] of the [keyData] must match with the [curve] specified in the parameter.

OR

The [keyData] should be generated for the same curve specified in the [curve] parameter.

jonasfj (approved):
nit suggestion

lib/src/webcrypto/webcrypto.ecdh.dart
jonasfj: Yeah, we could write either!

Issue #115, opened 2024-05-13
Consider using [!NOTE] blockquote for browser/platform compatibility notes


jonasfj:
In: https://pub.dev/documentation/webcrypto/latest/webcrypto/AesCtrSecretKey/encryptBytes.html

And a few other places we write: `**Remark** ...` to indicate that a particular browser or platform doesn't support a particular feature.

Instead of using this form, it'd be cool to use the [`[!NOTE]` format](https://github.com/orgs/community/discussions/16925).

**Example**

> [!NOTE]
> Remark Firefox does not implement counter rollover for AES-CTR correctly. Picking a sufficiently large `length` and using a `counter` that isn't filled with 0xff will likely avoid counter rollovers. See [bug 1803105](https://bugzilla.mozilla.org/show_bug.cgi?id=1803105) for details.

**Markdown**

```md
> [!NOTE]
> Remark Firefox does not implement counter rollover for AES-CTR correctly. Picking a sufficiently large `length` and using a `counter` that isn't filled with 0xff will likely avoid counter rollovers. See [bug 1803105](https://bugzilla.mozilla.org/show_bug.cgi?id=1803105) for details.
```

---------------------

I think this should work with recent versions of `dartdoc` too! :rocket: 

We should test it out locally first, and we should make sure to migrate all such remarks in one PR.

TropicolX:
Hi @jonasfj. Can I be assigned to this?

jonasfj:
Done,

jonasfj:
There is a thread about this package in the #packages channel on dart_community

https://dart.dev/community

Feel free to join and say hi.

TropicolX:
Alright, thanks!
Will do

TropicolX:
Hi @jonasfj,
How do I test out the changes locally? Can I get a preview UI similar to this: https://pub.dev/documentation/webcrypto/latest/webcrypto/AesCtrSecretKey/encryptBytes.html

jonasfj:
Run `dartdoc`

See: https://pub.dev/packages/dartdoc

TropicolX:
@jonasfj Changed all "Remark" notes to use the [!NOTE] format.

[...]

PR #126, opened 2024-05-24
Change all "Remark" notes to use the [!NOTE] blockquotes.

TropicolX:
To solve #115

### DIFF
--- a/example/pubspec.lock
+++ b/example/pubspec.lock
@@ -203,26 +203,26 @@ packages:
     dependency: transitive
     description:
       name: leak_tracker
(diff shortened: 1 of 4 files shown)


example/pubspec.lock
jonasfj, line 1: I think it's preferable to have such changes in a separate PR, explaining why we do this.

I'm guessing a newer Flutter version will cause this, no problem. But it's still ideally an independent PR.

jonasfj (approved):

example/pubspec.lock
TropicolX: Alright, noted! Should I undo the changes?

jonasfj (approved):

PR #127, opened 2024-05-28
Add Doc comments for EcdhPrivateKey Class

HamdaanAliQuatil:

PR #128, opened 2024-05-30
Document deriveBits in the EcdhPrivateKey Class

HamdaanAliQuatil:


HamdaanAliQuatil:
Split from #114 
The `@template` for the `@macro` used is defined in #127

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -38,6 +38,19 @@ abstract class EcdhPrivateKey {
     return impl.ecdhPrivateKey_generateKey(curve);
   }
 
+  /// Derives a shared secret from two ECDH key pairs using the private key from one pair 
+  /// and the public key from another.
```
jonasfj, line 42:
```suggestion
  /// Derive a shared secret from two ECDH key pairs using the private key from one pair 
  /// and the public key from another.
```

jonasfj (approved):

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -38,6 +38,19 @@ abstract class EcdhPrivateKey {
...
+  /// 
+  /// The shared secret is identical whether using A's private key and B's public key, 
+  /// or B's private key and A's public key, enabling secure key exchange between the 
+  /// two parties.
+  ///
+  /// [length] specifies the length of the derived secret in bits.
```
jonasfj, line 48: Should we make a list bullet point list with what lengths the different curves can provide:
```md
[length] specifies the length of the derived secret in bits. The maximum length depends on the [EllipticCurve] used in the keys:
...
 * [EllipticCurve.p521] can provide up to 528 bits.
```

or something like that.

jonasfj: We'd have to find out how much each curve can derive, and make sure we have tests :D

jonasfj: Actually, let's do this in a follow up issue.. That way we land now.

Landing now is always better :rofl:

jonasfj: Filed https://github.com/google/webcrypto.dart/issues/130

PR #131, opened 2024-05-31
Document exportPkcs8Key in the EcdhPrivateKey Class

HamdaanAliQuatil:


HamdaanAliQuatil:
Split from #114

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -204,6 +204,25 @@ abstract class EcdhPrivateKey {
   // https://tools.ietf.org/html/rfc6090#appendix-B
   Future<Uint8List> deriveBits(int length, EcdhPublicKey publicKey);
 
+  /// Export the [EcdhPrivateKey] as a [PKCS #8][1] key.
+  ///
+  /// Returns private key as a list of bytes.
```
jonasfj, line 209:
```suggestion
  /// Returns the DER encoding of the _PrivateKeyInfo_ structure specified in [RFC 5208][1] as a list of bytes.
```

jonasfj: That matches what we say in:

https://pub.dev/documentation/webcrypto/latest/webcrypto/RsassaPkcs1V15PrivateKey/exportPkcs8Key.html

And it lines up with what we say `importPkcs8Key` requires.

-----

Notably, I think it's nice to know that:
 * It's the _PrivateKeyInfo_ structure -- that way you know what to look at in the RFC.
 * It's DER encoded, and not BER, CER or some other ASN.1 variation, it's always DER.
 * It returns bytes, not something based64 encoded -- this is also indicated by the example that uses `package:pem` to wrap the output.

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -204,6 +204,25 @@ abstract class EcdhPrivateKey {
...
+  /// import 'package:pem/pem.dart';
+  /// import 'package:webcrypto/webcrypto.dart';
+  /// 
+  /// Future<void> main() async {
+  ///   // Alice generates a key-pair
+  ///   final kpA = await EcdhPrivateKey.generateKey(EllipticCurve.p256);
+  /// 
+  ///   // Export the private key (print it in same format as it was given).
+  ///   final exportedPkcs8Key = await kpA.privateKey.exportPkcs8Key();
+  ///   print(PemCodec(PemLabel.privateKey).encode(exportedPkcs8Key));
```
jonasfj, line 222: We don't need two parties, so let's not mention Alice or Bob.
```suggestion
  ///   // Generate a key-pair
  ///   final kp = await EcdhPrivateKey.generateKey(EllipticCurve.p256);
  /// 
  ///   // Export the private key.
  ///   final exportedPkcs8Key = await kp.privateKey.exportPkcs8Key();
  ///
  ///   // Private keys are often encoded as PEM.
  ///   // This encodes the key in base64 and wraps it with:
  ///   // '-----BEGIN PRIVATE KEY----'...
  ///   print(PemCodec(PemLabel.privateKey).encode(exportedPkcs8Key));
```

See: https://pub.dev/documentation/webcrypto/latest/webcrypto/RsassaPkcs1V15PrivateKey/exportPkcs8Key.html

jonasfj (changes requested):

jonasfj (approved):

PR #134, opened 2024-06-03
Add test coverage collection with coveralls

### LINKED ISSUE #133: Add test coverage collection with coveralls
See example setup in:
https://github.com/dart-lang/yaml_edit/blob/main/.github/workflows/test-package.yml

HamdaanAliQuatil:
Closes #133 

I've used the release `v2` in this collection (https://github.com/coverallsapp/github-action/releases).
It'd be interesting to know why [dart-lang](https://github.com/dart-lang/yaml_edit/blob/main/.github/workflows/test-package.yml) uses the specific commit and if we need to do the same.


.github/workflows/test.yml
```diff
@@ -136,3 +136,14 @@ jobs:
...
+  report-coverage:
+    needs: [linux, macos, windows, ios, android]
+    if: ${{ always() }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: coverallsapp/github-action@v2
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          parallel-finished: true
+
```
jonasfj, line 149: Something weird here? missing end of line?

.github/workflows/test.yml
```diff
@@ -136,3 +136,14 @@ jobs:
           api-level: 28
           working-directory: ./example
           script: flutter test integration_test/webcrypto_test.dart -d android
+
+  report-coverage:
+    needs: [linux, macos, windows, ios, android]
+    if: ${{ always() }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: coverallsapp/github-action@v2
```
jonasfj, line 145: We always use hashes rather than `@v2`. I think it's a security consideration, see:
https://opensource.google/documentation/reference/github/services

It's also nice because then action won't stop working because it somehow changed.

If we want to upgrade actions like this we can probably get dependabot to do it. Which we can also configure in the `.github/` folder.

.github/workflows/test.yml
```diff
@@ -136,3 +136,14 @@ jobs:
           api-level: 28
           working-directory: ./example
           script: flutter test integration_test/webcrypto_test.dart -d android
+
+  report-coverage:
```
jonasfj, line 140: We'll also need to update the other jobs too, right?

jonasfj:
I enabled coveralls.io for this repository.

We can add the following badge to the README too. Though it's not much fun until we've figured out how to get coverage numbers out of the various tests, combined them and submit them.

```md
[![Coverage Status](https://coveralls.io/repos/github/google/webcrypto.dart/badge.svg)](https://coveralls.io/github/google/webcrypto.dart)
```

It would also be cool to collect coverage numbers from integration tests, but that's probably hard to do, and I'm not sure it'll increase coverage much. Because it's the same tests we run :D

Probably just having coverage numbers for normal tests on Linux, Chrome and maybe Windows and Mac would be fine.

.github/workflows/test.yml
HamdaanAliQuatil: updated to the hash `643bc377ffa44ace6394b2b5d0d3950076de9f63` 🙌

.github/workflows/test.yml
HamdaanAliQuatil: maybe it was an indentation issue. edited the line. lmk if this works

.github/workflows/test.yml
```diff
@@ -25,9 +25,11 @@ jobs:
...
       - run: flutter pub get
       - run: flutter pub run webcrypto:setup
-      - run: flutter test
-      - run: flutter test --platform chrome
-      - run: xvfb-run flutter test integration_test/webcrypto_test.dart -d linux
+      - name: Run tests and collect coverage
+        run: |
+          flutter test --coverage
+          flutter test --platform chrome --coverage
+          xvfb-run flutter test integration_test/webcrypto_test.dart -d linux --coverage
```
jonasfj, line 32: I think we should keep them as different steps, so that it's easy to see which test failed. If a test actually fails.

It's also much easier if the display of the step we're running indicates what it is we're running.

That said, it's fine to exclude `--coverage` or `xvfb-run` from the display name.

Suggestion below:

```suggestion
      - name: flutter test
        run: flutter test --coverage
      - name: flutter test --platform chrome
        run: flutter test --platform chrome --coverage
      - name: flutter test integration_test/webcrypto_test.dart -d linux
        run: xvfb-run flutter test integration_test/webcrypto_test.dart -d linux --coverage
```

We should do similar for the other cases.

example/test/widget_test.dart
```diff
@@ -33,7 +33,7 @@ void main() {
     expect(
       find.byWidgetPredicate(
         (Widget widget) => widget is Text &&
-                           widget.data.startsWith('Running on:'),
+                           widget.data?.startsWith('Running on:'),
```
jonasfj, line 36:
```suggestion
                           widget.data?.startsWith('Running on:') ?? false,
```

jonasfj: This is probably why the tests fail :rofl:

coverage/lcov.info
jonasfj, line 1: let's not commit this file... maybe just add `/coverage/` to `.gitignore` ?

jonasfj:
Weird, that the tests seem to fail:
```
🎉 1433 tests passed.
Error: Process completed with exit code 1.
```

Do they also fail locally when running with `--coverage`? Like what is the exit code locally?

It seems like the test actually pass, but that the program still exits non-zero.
Maybe there is a bug somewhere. Maybe this is specific to `--platform chrome`, or?

Worth investigating for sure.

HamdaanAliQuatil:
> Weird, that the tests seem to fail:
> 
> ```
> 🎉 1433 tests passed.
> Error: Process completed with exit code 1.
> ```
> 
> Do they also fail locally when running with `--coverage`? Like what is the exit code locally?
> 
> It seems like the test actually pass, but that the program still exits non-zero. Maybe there is a bug somewhere. Maybe this is specific to `--platform chrome`, or?
> 
> Worth investigating for sure.

`--platform chrome` gives me a timeout error everytime. All tests passed in `flutter test --coverage` :
https://gist.github.com/HamdaanAliQuatil/77f9b09bba4929171c9fa11b3949250a

jonasfj:
> --platform chrome gives me a timeout error everytime. All tests passed in flutter test --coverage :

But it works on Github Actions. This might be windows specific, it might be specific to your setup. Feel free to debug it, you can try to see if `flutter pub run test --platform chrome` works, and you can try to do verbose mode, etc.

But it doesn't seem related to this issue.

coverage/lcov.info
HamdaanAliQuatil: added to `.gitignore`

HamdaanAliQuatil:
@jonasfj perhaps we don't need this now XD.
Should I close this?

jonasfj:
Yeah, let's close this.

I don't think we'll get much improvement by collecting coverage from iphone and android emulators, maybe a little, but I'm guessing not much.

We can explore it at some point in the future, but I think our immediate focus should be finishing the refactoring and write more tests :D

PR #135, opened 2024-06-16
Add tests for Import Key Exception

HamdaanAliQuatil:
- [x] Created optional param for importKeyException.
- [x] Added new Test Data with the importKeyException param.
- [x] Test passes if:
        1. Import Key Exception is not null in Test Data.
        2. And curve for the key in Test Data is not the same as the curve specified during import.
- [x] Test for deriveBits and subsequent tests should not proceed if the importKeyException is found.

### DIFF
--- a/lib/src/testing/utils/testrunner.dart
+++ b/lib/src/testing/utils/testrunner.dart
@@ -63,8 +63,9 @@ class _TestCase {
   // Parameters for key import (always required)
   final Map<String, dynamic>? importKeyParams;
 
-  // Parameters for import key exception (optional, if there is an exception)
-  final Map<String, dynamic>? importKeyException;
+  // If not `null`, then importing private or public key MUST throw
+  // an exception that contains this string.
+  final String? importKeyException;
 
   // Parameters for sign/verify (required, if there is a signature)
   final Map<String, dynamic>? signVerifyParams;
@@ -74,20 +75,6 @@ class _TestCase {
 
   // Parameters for deriveBits (required, if there is a derivedBits)
   final Map<String, dynamic>? deriveParams;
-
-  String? get pkcs8ImportKeyException {
-    if (importKeyException != null) {
-      return importKeyException?['pkcs8Exception'];
-    }
-    return null;
-  }
-
-  String? get jwkImportKeyException {
-    if (importKeyException != null) {
-      return importKeyException?['jwkException'];
-    }
-    return null;
-  }
 
   _TestCase(
     this.name, {
@@ -128,7 +115,7 @@ class _TestCase {
       derivedBits: _optionalBase64Decode(json['derivedBits']),
       derivedLength: json['derivedLength'] as int?,
       importKeyParams: _optionalStringMapDecode(json['importKeyParams']),
-      importKeyException: _optionalStringMapDecode(json['importKeyException']),
+      importKeyException: json['importKeyException'] as String?,
       signVerifyParams: _optionalStringMapDecode(json['signVerifyParams']),
       encryptDecryptParams:
           _optionalStringMapDecode(json['encryptDecryptParams']),
@@ -710,7 +697,14 @@ void _validateTestCase<PrivateKey, PublicKey>(
   check(c.importKeyParams != null);
   check((c.signVerifyParams != null) == (r._signBytes != null));
   check((c.encryptDecryptParams != null) == (r._encryptBytes != null));
-  check((c.deriveParams != null) == (r._deriveBits != null));
+
+  if (c.deriveParams != null) {
+    check((c.deriveParams != null) == (r._deriveBits != null));
+
+    if (r._deriveBits != null) {
+      check(c.derivedLength != null);
+    }
+  }
   if (c.signature != null) {
     check(r._signBytes != null);
   }
@@ -720,9 +714,6 @@ void _validateTestCase<PrivateKey, PublicKey>(
   if (c.derivedBits != null) {
     check(r._deriveBits != null);
   }
-  if (r._deriveBits != null) {
-    check(c.derivedLength != null);
-  }
 
   // Check that data matches the methods we have in the runner.
   check(r._importPrivateRawKey != null || c.privateRawKeyData == null);
@@ -731,6 +722,30 @@ void _validateTestCase<PrivateKey, PublicKey>(
   check(r._importPublicRawKey != null || c.publicRawKeyData == null);
   check(r._importPublicSpkiKey != null || c.publicSpkiKeyData == null);
   check(r._importPublicJsonWebKey != null || c.publicJsonWebKeyData == null);
+
+  if (c.generateKeyParams != null) {
+    check(c.importKeyException == null,
+        'importKeyException must be null when generateKeyParams is provided');
+  }
+
+  if (c.importKeyException != null) {
+    check(c.plaintext == null,
+        'plaintext must be null when importKeyException is provided');
+    check(c.signature == null,
+        'signature must be null when importKeyException is provided');
+    check(c.ciphertext == null,
+        'ciphertext must be null when importKeyException is provided');
+    check(c.derivedBits == null,
+        'derivedBits must be null when importKeyException is provided');
+    check(c.derivedLength == null,
+        'derivedLength must be null when importKeyException is provided');
+    check(c.signVerifyParams == null,
+        'signVerifyParams must be null when importKeyException is provided');
+    check(c.encryptDecryptParams == null,
+        'encryptDecryptParams must be null when importKeyException is provided');
+    check(c.deriveParams == null,
+        'deriveParams must be null when importKeyException is provided');
+  }
 }
 
 void _runTests<PrivateKey, PublicKey>(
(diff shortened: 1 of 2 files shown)


lib/src/testing/utils/testrunner.dart
```diff
@@ -63,6 +63,9 @@ class _TestCase {
   // Parameters for key import (always required)
   final Map<String, dynamic>? importKeyParams;
 
+  // Parameters for import key exception (optional, if there is an exception)
+  final Map<String, dynamic>? importKeyException;
```
jonasfj, line 67: Let's make this simpler.

```suggestion
  // If not `null`, then importing private or public key MUST throw
  // an exception that contains this string.
  final String? importKeyException;
```

Instead of making a test case that contains different strings for `pkcs8Exception` and `jwkException`, we just make two test cases. One that tests the exception when importing a PKCS8 key and on that tests it when importing a JWK key.

Notice that when specifying a test case, you don't have to specify both `privatePkcs8KeyData` and `privateJsonWebKeyData`, it's enough to specify one of them. So if we want to test different error messages (from different formats), we really should just make two test cases.

That way, we also avoid having an absolute explosion of properties on the test case objects.

jonasfj: We also need to update `_validateTestCase`

lib/src/testing/utils/testrunner.dart
```diff
@@ -746,6 +765,23 @@ void _runTests<PrivateKey, PublicKey>(
       publicKey = pair.publicKey;
       privateKey = pair.privateKey;
     });
+  } else if (c.importKeyException != null) {
+    test('pkcs8 import exception', () async {
+      try {
```
jonasfj, line 770: formatting of code is weird here. Please use `dart format` you can enable it in vscode with the Dart extension. I think it's even on by default.

lib/src/testing/utils/testrunner.dart
```diff
@@ -746,6 +765,23 @@ void _runTests<PrivateKey, PublicKey>(
...
+    }
+    });
+    test('jwk import exception', () async {
+      try {
+      await r._importPrivateJsonWebKey!(c.privateJsonWebKeyData!, {'curve': 'p-521'});
+      check(false, 'Expected an exception for P-512 import');
+    } catch (e) {
+      check(e.toString().contains(c.jwkImportKeyException!));
+    }
+    });
```
jonasfj, line 784: If `importKeyException` is not `null`, then we can test that importing public and private keys fail, but all the other tests we can't do, right?

So maybe it's fine to just do a `return;` here? That way we don't have to add `&& c.pkcs8ImportKeyException == null` to all the other test cases.

jonasfj: We should also when doing this, make sure that in `_validateTestCase` we validate that the test case doesn't have values that won't be used.

In particular we need to validate that:
 * if `generateKeyParams` is not null, then `importKeyException` must be `null`!
 * if `importKeyException` is not null, then
   * All the other properties that won't be used must be `null`.
     * This includes `plaintext`, `signature`, `ciphertext`, `derivedBits`, `derivedLength` and more.
       (If the tests don't use the values, they MUST not be specified, otherwise the test case will be hard to read and understand)
     * Probably also includes `signVerifyParams`, etc. since again, we can't use them for anything, so they shouldn't be specified.
   * At-least one of:
     * `publicRawKeyData`, `publicSpkiKeyData`, `publicJsonWebKeyData`, `privateRawKeyData`, `privatePkcs8KeyData`, or `privateJsonWebKeyData` must be specified.
     * Note: that we probably don't need all of them to be specified.
     * Note: it's probably okay if there is no private key or no public key, just so long as there is at-least ONE of these properties provided. Normally, when `importKeyException == null`, this is ofcourse not allowed.

lib/src/testing/utils/testrunner.dart
HamdaanAliQuatil: Updated with one string `importKeyException`. Made necessary changes to `_validateTestCase`

lib/src/testing/utils/testrunner.dart
HamdaanAliQuatil: Tried `dart format`, but nothing seemed to change on this line. A few other parts were formatted though.

lib/src/testing/utils/testrunner.dart
HamdaanAliQuatil: Added a `return;` in this section. Removed recurring `&& c.pkcs8ImportKeyException == null`
Updated `_validateTestCase` with the above checks

lib/src/testing/utils/testrunner.dart
```diff
@@ -691,7 +697,14 @@ void _validateTestCase<PrivateKey, PublicKey>(
...
   check((c.encryptDecryptParams != null) == (r._encryptBytes != null));
-  check((c.deriveParams != null) == (r._deriveBits != null));
+
+  if (c.deriveParams != null) {
+    check((c.deriveParams != null) == (r._deriveBits != null));
+
+    if (r._deriveBits != null) {
+      check(c.derivedLength != null);
+    }
+  }
```
jonasfj, line 707: Is this unrelated to this change, if so let's make it a different PR

lib/src/testing/utils/testrunner.dart
```diff
@@ -746,6 +780,30 @@ void _runTests<PrivateKey, PublicKey>(
...
+      test('jwk import exception', () async {
+        try {
+          await r._importPrivateJsonWebKey!(
+              c.privateJsonWebKeyData!, {'curve': 'p-521'});
+          check(false, 'Expected an exception for P-512 import');
+        } catch (e) {
+          check(e.toString().contains(c.importKeyException!));
+        }
+      });
+    }
```
jonasfj, line 805: We need a test case for each kind of key possible:
```
  final List<int>? privateRawKeyData;
  final List<int>? privatePkcs8KeyData;
  final Map<String, dynamic>? privateJsonWebKeyData;
  final List<int>? publicRawKeyData;
  final List<int>? publicSpkiKeyData;
  final Map<String, dynamic>? publicJsonWebKeyData;
```

Also I'd suggest naming the test cases like:
```dart
if (c.publicRawKeyData != null) {
  test('importPublicRawKey() throws', () async {
    ...
  });
}
if (c.publicJsonWebKeyData != null) {
  test('importPublicJsonWebKey() throws', () async {
    ...
  });
}
...

// This test case is testing failure to import a key
// further tests will not be conducted, since they don't
// make sense after failing to import a key.
return;
```

lib/src/testing/webcrypto/ecdh.dart
```diff
@@ -111,6 +111,39 @@ final _testData = [
...
     "deriveParams": {}
   },
+  {
+    "name": "generated on boringssl/linux (pkcs8 import key exception) at 2020-01-22T23:24:34",
+    "privatePkcs8KeyData":
+        "MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQg3aTiZ7odKAODYk4BpZlzulBCB/BptmxjtvrzyXI71UyhRANCAATl0GVa8O1sXXf2NV5qGJ/9/Vq8PVWCZuezADa1F0Vr2TaB8BseZIW+rhmEmLC2FfCdxj9NmLp00SilRTm40Hxm",
+    "publicRawKeyData":
+        "BHiIXxrwhM92v4ueDrj3x1JJY4uS+II/IJPjqMvaKj/QfoOllnEkrnaOW1owBYRBMnP0pPouPkqbVfPACMUsfKs=",
+    "publicSpkiKeyData":
+        "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEeIhfGvCEz3a/i54OuPfHUklji5L4gj8gk+Ooy9oqP9B+g6WWcSSudo5bWjAFhEEyc/Sk+i4+SptV88AIxSx8qw==",
```
jonasfj, line 121: We should change the `_validateTestCase` further.

Right now a test case is require to have one of:
 * import key parameters,
 * if symmetric has private key (and not public key)
 * if not symmetric has both:
   * a private key, and,
   * a public key.

We should want that if `importKeyException` then it is not necessary to have both private and public key.

Because, in some cases you may want to check an exception that can only be produced when trying to import a private key (or trying to import a public key).

Example: trying to import a public key will never complain about `d` being missing from the JWT.

lib/src/testing/webcrypto/ecdh.dart
```diff
@@ -111,6 +111,39 @@ final _testData = [
     "importKeyParams": {"curve": "p-256"},
     "deriveParams": {}
   },
+  {
+    "name": "generated on boringssl/linux (pkcs8 import key exception) at 2020-01-22T23:24:34",
```
jonasfj, line 115: I think we can do a better "name" here.

For exceptions, we shouldn't really care where the test case was generated.
Because it's NOT valid, you probably did something to it to make it invalid! :rofl: 

So better make the test case name say: "import incorrect curve" or something like that.

lib/src/testing/webcrypto/ecdh.dart
```diff
@@ -111,6 +111,39 @@ final _testData = [
...
+      "y": "NoHwGx5khb6uGYSYsLYV8J3GP02YunTRKKVFObjQfGY",
+      "d": "3aTiZ7odKAODYk4BpZlzulBCB_BptmxjtvrzyXI71Uw"
+    },
+    "publicRawKeyData":
+        "BHiIXxrwhM92v4ueDrj3x1JJY4uS+II/IJPjqMvaKj/QfoOllnEkrnaOW1owBYRBMnP0pPouPkqbVfPACMUsfKs=",
+    "publicSpkiKeyData":
+        "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEeIhfGvCEz3a/i54OuPfHUklji5L4gj8gk+Ooy9oqP9B+g6WWcSSudo5bWjAFhEEyc/Sk+i4+SptV88AIxSx8qw==",
+    "publicJsonWebKeyData": {
+      "kty": "EC",
+      "crv": "P-256",
```
jonasfj, line 140: This curve doesn't look wrong?

mosuem:
Hi @HamdaanAliQuatil , just checking in on this old PR - is this something you still intend to merge?

HamdaanAliQuatil:
no @mosuem this is quite stall. and we need to rethink our entire test stuite again anyways. will create new prs after a new design is finalized.

Issue #105, opened 2024-05-06
Avoid @sealed annotation, using `final class` for everything


jonasfj:
We should use `final class` for everything in the public API.

**Motiviation**:
* This package is **not** attempting to offer abstractions over cryptographic APIs.
* We don't want others to implement the types we offer.
* We also want the option to add more method to the types we offer in the future, without breaking API compatibility.

In practice, we might have to do something that we export from `src/webcrypto/webcrypto.dart` and something we only export to other libraries inside this package.

```dart
// This is exported from lib/webcrypto.dart
// This means that in lib/webcrypto we have to use "show"
// to avoid exporting KeyPairBase.
// Example:
//   export src/webcrypto.dart show KeyPair;
abstract final class KeyPair<S, T> {
  /// Private key for [publicKey].
  S get privateKey;

  /// Public key matching [privateKey].
  T get publicKey;
}

// This is not exported from lib/webcrypto.dart
// but implementations for different platforms can import this from
// src/webcrypto/webcrypto.dart
// and extend it inorder to implement KeyPair
abstract base class KeyPairBase<S, T> extends KeyPair<S, T> {}
```

devenderbutani21:
Hello! I would like to work in this issue.

jonasfj:
Assigned to you.

This is mostly a bit of refactoring and making sure the public API is pretty, while also making sure people accessing it can't `implement` our classes :D

PR #141, opened 2024-07-24
Update AesCbcSecretKey Class

HamdaanAliQuatil:


HamdaanAliQuatil:
part of #105

lib/src/impl_stub/impl_stub.dart
```diff
@@ -0,0 +1,11 @@
...
+
+import 'dart:typed_data';
+import 'dart:async';
+
+import '../impl_stub.dart'
+    if (dart.library.ffi) '../impl_ffi/impl_ffi.dart'
+    if (dart.library.js) '../impl_js/impl_js.dart' as impl;
+    
+
+part 'impl_stub.aescbc.dart';
```
jonasfj, line 11: please end files with a line break.

See https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files

lib/src/impl_stub/impl_stub.aescbc.dart
```diff
@@ -0,0 +1,22 @@
+part of 'impl_stub.dart';
+
+abstract class AesCbcSecretKeyImpl {
```
jonasfj, line 3:
```suggestion
abstract base class AesCbcSecretKeyImpl {
```
or
```suggestion
abstract interface class AesCbcSecretKeyImpl {
```

lib/src/impl_stub/impl_stub.dart
```diff
@@ -0,0 +1,11 @@
+library impl_stub;
+
+import 'dart:typed_data';
+import 'dart:async';
+
+import '../impl_stub.dart'
+    if (dart.library.ffi) '../impl_ffi/impl_ffi.dart'
+    if (dart.library.js) '../impl_js/impl_js.dart' as impl;
```
jonasfj, line 8: okay, I forgot we had `lib/src/impl_stub.dart` already. So `lib/src/impl_stub/...` is perhaps a bad name.

I'll try to outline what we have:

 * `lib/webcrypto.dart` exports public classes from `lib/src/webcrypto/webcrypto.dart`, all classes exported should be `final`.
 * `lib/src/webcrypto/webcrypto.dart` contains the implementation of the public classes (these will be `final`).
 * `lib/src/impl_stub.dart`, contains a "stub" platform implementation.
    This implements everything that a platform implementation must provide, it just does it in the trivial way where every method throws `UnimplementedError`.
    This platform implementation is kind of useless, but it's technically a correct implementation :rofl: 
 * `lib/src/impl_js/impl_js.dart`, contains a platform implementation using js.
    This library has the same interface (top-level methods) as `lib/src/impl_stub.dart`.
 * `lib/src/impl_js/impl_ffi.dart`, contains a platform implementation using ffi.
    This library has the same interface (top-level methods) as `lib/src/impl_stub.dart`.


So `lib/src/webcrypto/webcrypto.dart` needs to import either:
 * `lib/src/impl_stub.dart`,
 * `lib/src/impl_js/impl_js.dart`, or.
 * `lib/src/impl_js/impl_ffi.dart`.

They all have the same top-level methods and interface, so we can use configurable imports.

------------------------

What we're trying to do in `lib/src/impl_stub/impl_stub.dart` is to create a library the specifies the interfaces that platform libraries should implement. That is interfaces that should be implemented by:
 * `lib/src/impl_stub.dart`,
 * `lib/src/impl_js/impl_js.dart`, and,
 * `lib/src/impl_js/impl_ffi.dart`.

Perhaps calling it `lib/src/impl_stub/impl_stub.dart` is a mistake. Maybe, it's better to call it `impl_interface` ?

After all we want an import-graph where:
 * `lib/webcrypto.dart` exports:
 * `lib/src/webcrypto/webcrypto.dart` which imports, one of:
   * `lib/src/impl_stub.dart`,
   * `lib/src/impl_js/impl_js.dart`, or
   * `lib/src/impl_js/impl_ffi.dart`.
   Each of which imports:
 * `lib/src/impl_interface/impl_interface.dart`.


-------

Maybe, (this is just an idea) we should make:
```dart
# lib/src/impl_interface/impl_interface.dart

/// A platform implementation must provide a public top-level constant implementation
/// of [WebCryptoImpl] called `webCryptImpl`.
abstract interface class WebCryptoImpl {
  StaticAesCbcSecretKeyImpl get aesCbcSecretKey;
}

# lib/src/impl_interface/impl_interface.aescbc.dart

abstract interface class StaticAesCbcSecretKeyImpl {
  Future<AesCbcSecretKeyImpl> importRawKey(List<int> keyData);
  Future<AesCbcSecretKeyImpl> importJsonWebKey(Map<String, dynamic> jwk);
  Future<AesCbcSecretKeyImpl> generateKey(int length);
}

abstract interface class AesCbcSecretKeyImpl {
  Future<Uint8List> encryptBytes(List<int> data, List<int> iv);
  Future<Uint8List> decryptBytes(List<int> data, List<int> iv);
  Stream<Uint8List> encryptStream(Stream<List<int>> data, List<int> iv);
  Stream<Uint8List> decryptStream(Stream<List<int>> data, List<int> iv);
  Future<Uint8List> exportRawKey();
  Future<Map<String, dynamic>> exportJsonWebKey();
}

# lib/src/impl_stub/impl_stub.dart  // this is now replacing lib/src/impl_stub.dart

const webCryptImpl = _WebCryptoImpl();

final class _WebCryptoImpl implememts WebCryptoImpl {
  const _WebCryptoImpl();

  final aesCbcSecretKey = _StaticAesCbcSecretKeyImpl();
}

# lib/src/impl_stub/impl_stub.aescbc.dart

final class _StaticAesCbcSecretKeyImpl implements StaticAesCbcSecretKeyImpl {
  const _StaticAesCbcSecretKeyImpl();

  @override
  Future<AesCbcSecretKeyImpl> importRawKey(List<int> keyData) =>
    throw UnimplementedError('not implemented');

  @override
  Future<AesCbcSecretKeyImpl> importJsonWebKey(Map<String, dynamic> jwk) =>
    throw UnimplementedError('not implemented');

  @override
  Future<AesCbcSecretKeyImpl> generateKey(int length) =>
    throw UnimplementedError('not implemented');
}

# lib/src/impl_js/impl_js.dart

const webCryptImpl = _WebCryptoImpl();

final class _WebCryptoImpl implememts WebCryptoImpl {
  const _WebCryptoImpl();

  final aesCbcSecretKey = _StaticAesCbcSecretKeyImpl();
}

# lib/src/impl_js/impl_js.aescbc.dart

final class _StaticAesCbcSecretKeyImpl implements StaticAesCbcSecretKeyImpl {
  const _StaticAesCbcSecretKeyImpl();

  @override
  Future<AesCbcSecretKeyImpl> importRawKey(List<int> keyData) async {
     ...
     
     return _AesCbcSecretKeyImpl(...)
  }

  @override
  Future<AesCbcSecretKeyImpl> importJsonWebKey(Map<String, dynamic> jwk) => ...

  @override
  Future<AesCbcSecretKeyImpl> generateKey(int length) => ...
}

final class _AesCbcSecretKeyImpl implements AesCbcSecretKeyImpl {
  ...
}

# lib/src/impl_ffi/impl_ffi.dart         // similar to impl_ffi.dart
# lib/src/impl_ffi/impl_ffi.aescbc.dart  // similar to impl_ffi.aescbc.dart

# lib/src/webcrypto/webcrypto.dart

import '../impl_interface/impl_interface.dart'
import '../impl_stub/impl_stub.dart'
    if (dart.library.ffi) '../impl_ffi/impl_ffi.dart'
    if (dart.library.js) '../impl_js/impl_js.dart';

# lib/src/webcrypto/webcrypto.aescbc.dart

final class AesCbcSecretKey {
  final AesCbcSecretKeyImpl _impl;

  AesCbcSecretKey._(this._impl); 
  
  static Future<AesCbcSecretKey> importRawKey(List<int> keyData) async {
    final impl = await webCryptImpl.aesCbcSecretKey.importRawKey(keyData);
    return AesCbcSecretKey._(impl);
  }

  ...

  Future<Uint8List> encryptBytes(List<int> data, List<int> iv) =>
      _impl.encryptBytes(data, iv);
  ...
}
```

Maybe, this is crazy overkill... it certainly could be a bit slower due to all the indirection, but since all the indirection is through constants, it might be not affect performance.

If we did this, then the top-level constant `webCryptImpl` is the only thing exported from:
 * `lib/src/impl_stub.dart`,
 * `lib/src/impl_js/impl_js.dart`, and
 * `lib/src/impl_js/impl_ffi.dart`.

And then the type checker will do a pretty decent job at making sure all implementations actually provide a `webCryptImpl` that correctly implements `WebCryptoImpl`.

So we don't end up relying so much on all these implementations providing the same top-level methods, because they just provide one top-level constant, that implements all the methods.

If we were really paranoid we could even do something like:
``` 
# lib/src/webcrypto/webcrypto.aescbc.dart

final class AesCbcSecretKey {
  final AesCbcSecretKeyImpl _impl;

  AesCbcSecretKey._(this._impl); 
  
  static Future<AesCbcSecretKey> importRawKey(List<int> keyData) async {
    assert(webCryptImpl is WebCryptoImpl); // Compiler should hopefully (who knows) be smart enough to compile this away if it's true, since webCryptImpl is a constant
    final impl = await webCryptImpl.aesCbcSecretKey.importRawKey(keyData);
    return AesCbcSecretKey._(impl);
  }

  ...

  Future<Uint8List> encryptBytes(List<int> data, List<int> iv) =>
      _impl.encryptBytes(data, iv);
  ...
}
```

lib/src/impl_stub/impl_stub.dart
HamdaanAliQuatil: Updated! thanks 🙌

lib/src/impl_stub/impl_stub.aescbc.dart
HamdaanAliQuatil: updated to `abstract interface`

lib/src/impl_stub/impl_stub.dart
HamdaanAliQuatil: Updated to the new architecture. Removed relevant methods from `src/impl_stub.dart`. Will phase out this file sequentially as all classes are refactored 🚀

lib/src/impl_interface/impl_interface.dart
```diff
@@ -0,0 +1,11 @@
+library impl_stub;
```
jonasfj, line 1: license headers :D

lib/src/impl_js/impl_js.aescbc.dart
```diff
@@ -28,28 +28,47 @@ Future<AesCbcSecretKey> aesCbc_importRawKey(List<int> keyData) async {
...
+    return await aesCbc_importJsonWebKey(jwk);
+  }
+
+  @override
+  Future<AesCbcSecretKeyImpl> generateKey(int length) async {
+    return await aesCbc_generateKey(length);
+  }
+}
+
+class _AesCbcSecretKeyImpl implements AesCbcSecretKeyImpl {
```
jonasfj, line 69:
```suggestion
final class _AesCbcSecretKeyImpl implements AesCbcSecretKeyImpl {
```

lib/src/impl_js/impl_js.dart
```diff
@@ -53,3 +55,12 @@ class _KeyPair<S, T> implements KeyPair<S, T> {
 
   _KeyPair({required this.privateKey, required this.publicKey});
 }
+
+const webCryptImpl = _WebCryptoImpl();
```
jonasfj, line 59: avoid leaking the type, not that it really matters, it's just prettier in my head :rofl: 
```suggestion
const WebCryptoImpl webCryptImpl = _WebCryptoImpl();
```

lib/src/impl_stub/impl_stub.dart
```diff
@@ -0,0 +1,14 @@
+library webcrypto.impl_stub;
+
+import 'package:webcrypto/src/impl_interface/impl_interface.dart';
+
+part 'impl_stub.aescbc.dart';
+
+const webCryptImpl = _WebCryptoImpl();
```
jonasfj, line 7: Again, not that it really matters, but now that static interface to each library is exactly the same :rofl: 
```suggestion
const WebCryptoImpl webCryptImpl = _WebCryptoImpl();
```

lib/src/webcrypto/webcrypto.dart
```diff
@@ -24,9 +24,13 @@ import 'package:meta/meta.dart';
 import 'dart:convert';
 import 'dart:async';
 import 'dart:typed_data';
+import '../impl_interface/impl_interface.dart';
 import '../impl_stub.dart'
     if (dart.library.ffi) '../impl_ffi/impl_ffi.dart'
     if (dart.library.js) '../impl_js/impl_js.dart' as impl;
+import '../impl_stub/impl_stub.dart'
+    if (dart.library.ffi) '../impl_ffi/impl_ffi.dart'
+    if (dart.library.js) '../impl_js/impl_js.dart';
```
jonasfj, line 33:
```suggestion
import '../impl_stub/impl_stub.dart'
    if (dart.library.ffi) '../impl_ffi/impl_ffi.dart'
    if (dart.library.js) '../impl_js/impl_js.dart' show webCryptImpl;
```

No matter which of these 3 libraries we import we only want to see the `webCryptImpl` symbol. And we've migrated all implementations then this really should be the only thing exported by all these 3 libraries.

This is actually pretty neat, it makes it much harder to accidentally export something from `impl_stub.dart` and then forget to export it from `impl_js.dart` (which won't be checked by `dart analyze`). The analyzer will only look at the first configurable import option.

jonasfj:
I think we should consider flagging this as Ready for review, and then land it, and move one primitive at the time to the new architecture.

Otherwise, we'll be creating a huge PR that'll be nearly impossible to review.

lib/src/impl_interface/impl_interface.dart
HamdaanAliQuatil: added headers

jonasfj (approved):

jonasfj:
Landed as part of https://github.com/google/webcrypto.dart/pull/142 @HamdaanAliQuatil correct me if I'm wrong.

PR #144, opened 2024-08-17
Refactor HmacSecretKey Class

HamdaanAliQuatil:


HamdaanAliQuatil:
part of #105

lib/src/impl_stub.dart
```diff
@@ -46,22 +46,6 @@ const Hash sha384 = _UnimplementedHash();
 const Hash sha512 = _UnimplementedHash();
 
 //---------------------- HMAC
```
jonasfj, line 48:
```suggestion
```

lib/src/impl_stub/impl_stub.hmac.dart
```diff
@@ -0,0 +1,34 @@
...
+// limitations under the License.
+
+part of 'impl_stub.dart';
+
+final class _StaticHmacSecretKeyImpl implements StaticHmacSecretKeyImpl {
+  const _StaticHmacSecretKeyImpl();
+
+  @override
+  Future<HmacSecretKeyImpl> importRawKey(List<int> keyData, Hash hash, {int? length}) {
+    throw UnimplementedError();
```
jonasfj, line 22: Can we throw `throw UnimplementedError('Not implemented');`

Or even just make a `final _notImplemented = UnimplementedError('Not implemented');` inside `impl_stub.dart`, then we can use that in all the stubs.

Perhaps this is a follow up PR :D

lib/src/webcrypto/webcrypto.hmac.dart
```diff
@@ -219,7 +230,7 @@ abstract class HmacSecretKey {
   /// instead, these methods computes a signature and does a
   /// fixed-time comparison.
   /// {@endtemplate}
-  Future<Uint8List> signBytes(List<int> data);
+  Future<Uint8List> signBytes(List<int> data) => _impl.signBytes(data);
```
jonasfj, line 233: We should probably do:
```suggestion
  Future<Uint8List> signBytes(List<int> data) async => await _impl.signBytes(data);
```

It can create different stack traces

jonasfj (approved):

lib/src/impl_stub/impl_stub.hmac.dart
HamdaanAliQuatil: Updated to `throw UnimplementedError('Not implemented');` as is in other files. Missed it here.

> Or even just make a final _notImplemented = UnimplementedError('Not implemented'); inside impl_stub.dart, then we can use that in all the stubs.

Will update to this structure in a follow up PR

jonasfj:
@HamdaanAliQuatil please rebase this, so we can run tests again, hopefully they'll pass :D

HamdaanAliQuatil:
> @HamdaanAliQuatil please rebase this, so we can run tests again, hopefully they'll pass :D

@jonasfj Done. The tests still didn't pass

PR #145, opened 2024-08-19
Refactor EcdhPublicKey and EcdhPrivateKey Class

HamdaanAliQuatil:


HamdaanAliQuatil:
Part of #105

lib/src/impl_interface/impl_interface.dart
```diff
@@ -17,9 +17,12 @@ library impl_stub;
 import 'dart:typed_data';
 import 'dart:async';
 
+import '../webcrypto/webcrypto.dart';
```
jonasfj, line 20: What we don't want this. That creates cyclic imports. It becomes messy, let's avoid that.

We can just make a `KeyPairImpl`, or fully move `KeyPair` into `impl_interface.dart`.

HamdaanAliQuatil: Yes, this makes sense. tysm. modifying accordingly

lib/src/impl_interface/impl_interface.ecdh.dart
```diff
@@ -0,0 +1,39 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+part of 'impl_interface.dart';
+
+abstract interface class StaticEcdhPrivateKeyImpl {
+  Future<EcdhPrivateKeyImpl> importPkcs8Key(List<int> keyData, EllipticCurve curve);
+  Future<EcdhPrivateKeyImpl> importJsonWebKey(Map<String, dynamic> jwk, EllipticCurve curve);
+  Future<KeyPair<EcdhPrivateKeyImpl, EcdhPublicKeyImpl>> generateKey(EllipticCurve curve);
```
jonasfj, line 20: We could do this right?
```suggestion
  Future<(EcdhPrivateKeyImpl, EcdhPublicKeyImpl)> generateKey(EllipticCurve curve);
```

lib/src/impl_interface/impl_interface.dart
jonasfj: Or we can just do as proposed in:
https://github.com/google/webcrypto.dart/pull/145/files#r1804350232

Specifcally, we do:
```
Future<(EcdhPrivateKeyImpl, EcdhPublicKeyImpl)> generateKey(EllipticCurve curve);
```

Then we can still make it into a `KeyPair` in `webcrypto/webcrypto.dart` where we have to wrap `EcdhPrivateKeyImpl` into `EcdhPrivateKey`.

jonasfj:
hmm, not sure my merge was correct, I might have created more problems than I solved :see_no_evil:

jonasfj (approved):

PR #155, opened 2024-09-27
Disable chromedriver tests on macos

jonasfj:
This is effectively integration tests for Chrome on MacOS. These tests shouldn't really be necessary, since we tests chrome compatibility in multiple ways.

For unknown reasons these tests are unreliable on MacOS. They were constantly timing out for no good reason.


HamdaanAliQuatil (approved):

HamdaanAliQuatil (approved):
LGTM ✅

szakarias (approved):

PR #160, opened 2024-10-18
Apply dart format

jonasfj:
Surprised we never had a test that to ensure `dart format` was applied, I assumed it was a given.


HamdaanAliQuatil (approved):
LGTM ✅

PR #167, opened 2024-10-23
Fix typo in test.yml

jonasfj:
Couldn't get the tests to run, so I just assumed it was a fluke. Only when I landed on `master` it was obvious that I had malformed YAML :see_no_evil:


HamdaanAliQuatil (approved):
LGTM ✅

PR #171, opened 2024-10-25
Refactor Hash Class

HamdaanAliQuatil:


HamdaanAliQuatil:
Part of #105

lib/src/impl_ffi/impl_ffi.digest.dart
```diff
@@ -62,39 +62,104 @@ abstract class _Hash implements Hash {
...
+    } else if (hash == sha256) {
+      return 'HS256';
+    } else if (hash == sha384) {
+      return 'HS384';
+    } else if (hash == sha512) {
+      return 'HS512';
+    } else {
+      throw UnsupportedError('hash is not supported');
+    }
+  }
```
jonasfj, line 79: Not sure where to start:
 1. This shouldn't be declared in `impl_interface.digest.dart` because this is only used inside `impl_ffi/`, right? (or am I wrong?)
 2. Why does it take an argument? :rofl: It can just operate on `this`. If you wanted it to take an argument, it should have been a static method.
 3. Why not make it abstract, and then implement it on the classes?
```suggestion
  String get hmacJwkAlg;
```

lib/src/impl_ffi/impl_ffi.digest.dart
```diff
@@ -62,39 +62,104 @@ abstract class _Hash implements Hash {
...
 }
 
-class _Sha1 extends _Hash {
+final class _Sha1 extends _HashImpl {
   const _Sha1();
 
   @override
   ffi.Pointer<EVP_MD> Function() get _algorithm => ssl.EVP_sha1;
 }
```
jonasfj, line 132:
```suggestion
final class _Sha1 extends _HashImpl {
  const _Sha1();
  
  @override
  String get hmacJwkAlg => 'HS1';

  @override
  ffi.Pointer<EVP_MD> Function() get _algorithm => ssl.EVP_sha1;
}
```

This is actually a bit better too than using the `if-elseif`, if someone only uses `sha256` then the other classes can be omitted from the binary (by tree shaking). So the final binary will be a bit smaller! :tada:

lib/src/impl_interface/impl_interface.digest.dart
```diff
@@ -0,0 +1,26 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+part of 'impl_interface.dart';
+
+abstract interface class HashImpl {
+  HashImpl(HashImpl impl);
```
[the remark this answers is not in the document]
HamdaanAliQuatil: I believe this was an oversight. This should not have slipped. Removed this and other abstract methods as they now have corresponding getters with the same names.

lib/src/impl_ffi/impl_ffi.digest.dart
```diff
@@ -14,17 +14,17 @@
...
+  factory _HashImpl.fromHash(HashImpl hash) {
+    if (hash is _HashImpl) {
       return hash;
     }
     throw ArgumentError.value(
       hash,
       'hash',
-      'Custom implementations of Hash is not supported',
+      'Custom implementations of HashImpl is not supported',
     );
```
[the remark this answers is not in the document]
HamdaanAliQuatil: Noted. Modified to `AssertionError`. 
Yeah, let's keep the method. I believe `_HashImpl.fromHash` provides a better developer experience and saves us from casting everywhere.

lib/src/impl_ffi/impl_ffi.hmac.dart
```diff
@@ -74,7 +57,7 @@ Future<HmacSecretKeyImpl> hmacSecretKey_importJsonWebKey(
   checkJwk(k.kty == 'oct', 'kty', 'must be "oct"');
   checkJwk(k.k != null, 'k', 'must be present');
   checkJwk(k.use == null || k.use == 'sig', 'use', 'must be "sig", if present');
-  final expectedAlg = _hmacJwkAlgFromHash(h);
+  final expectedAlg = _HashImpl.fromHash(h).hmacJwkAlg(h);
```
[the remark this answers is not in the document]
HamdaanAliQuatil: Now I see what I did 🙈 Fixed this with getters.

lib/src/impl_ffi/impl_ffi.digest.dart
HamdaanAliQuatil: Yup, it's only being used in `impl_ffi/`. I did write some vestigial functions in `impl/js`. Performed a cleanup now. All tests pass on local.

lib/src/impl_js/impl_js.digest.dart
```diff
@@ -14,9 +14,9 @@
 
 part of 'impl_js.dart';
 
-class _Hash implements Hash {
+class _HashImpl implements HashImpl {
```
jonasfj, line 17:
```suggestion
final class _HashImpl implements HashImpl {
```

People shouldn't be importing from `src/` and we shouldn't be subclassing it ourselves.

But if we make it `final` we make it clear to future us, that we have decided not to subclass `_HashImpl`.

lib/src/webcrypto/webcrypto.digest.dart
```diff
@@ -30,9 +30,11 @@ part of 'webcrypto.dart';
...
 ///
 /// [1]: https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-57pt1r5.pdf
-@sealed
-abstract class Hash {
-  Hash._(); // keep the constructor private.
+final class Hash {
+  final HashImpl _impl;
+
+  Hash._(this._impl); // keep the constructor private.
+  HashImpl get impl => _impl;
```
jonasfj, line 37: This shouldn't be a public property, hopefully we don't need that anywhere.
```suggestion
```

lib/src/webcrypto/webcrypto.digest.dart
```diff
@@ -98,7 +101,7 @@ abstract class Hash {
   /// ```
   ///
   /// [1]: https://doi.org/10.6028/NIST.FIPS.180-4
-  static const Hash sha1 = impl.sha1;
+  static Hash sha1 = Hash._(webCryptImpl.sha1);
```
jonasfj, line 104: This is a change in the API, and it could be breaking to people who rely on these being constants.

jonasfj: I think we could make them `const` with a tweak like:

```dart
abstract final class Hash {
  const Hash._();
  HashImpl get _impl;
  
  Future<Uint8List> digestBytes(List<int> data) => _impl.digestBytes(data);
  Future<Uint8List> digestStream(Stream<List<int>> data) =>
      _impl.digestStream(data);

  static const Hash sha256 = _Sha256();
}

final class _Sha256 extends Hash {
  const _Sha256() : super._();
  
  @override
  HashImpl get _impl => HashImpl();
}
```

lib/src/testing/webcrypto/digest.dart
```diff
@@ -63,7 +63,7 @@ List<({String name, Future<void> Function() test})> tests() {
   });
 
   test('SHA-1: libsum', () async {
-    const h = Hash.sha1;
+    final h = Hash.sha1;
```
jonasfj, line 66: We won't need this change if we make `Hash.sha1` `const` again. See other comment.

lib/src/impl_js/impl_js.digest.dart
```diff
@@ -35,14 +35,9 @@ class _Hash implements Hash {
...
-const Hash sha512 = _Hash('SHA-512');
-
 /// Get the algorithm from [hash] or throw an [ArgumentError].
-String _getHashAlgorithm(Hash hash) {
-  if (hash is _Hash) {
+String _getHashAlgorithm(HashImpl hash) {
+  if (hash is _HashImpl) {
     return hash._algorithm;
   }
   throw ArgumentError.value(
```
jonasfj, line 43: Let's also make this an `AssertionError`, because it shouldn't be possible, right?

lib/src/impl_ffi/impl_ffi.digest.dart
```diff
@@ -14,18 +14,15 @@
...
+    if (hash is _HashImpl) {
       return hash;
     }
-    throw ArgumentError.value(
-      hash,
-      'hash',
-      'Custom implementations of Hash is not supported',
-    );
+    throw AssertionError(
+        'Internal invariant violated: Custom implementations of HashImpl are not supported.');
```
jonasfj, line 25:
```suggestion
        'Custom implementations of HashImpl are not supported.');
```
An `AssertionError` is always an internal invariant that's violated.

Note. This is the same error we have in `_getHashAlgorithm` in `impl_js`.

lib/src/impl_ffi/impl_ffi.digest.dart
```diff
@@ -62,39 +59,100 @@ abstract class _Hash implements Hash {
...
     });
   }
+
+  String get hmacJwkAlg;
+
+  String get rsaOaepJwkAlg;
+
+  String get rsaPssJwkAlg;
+
+  String get rsassaPkcs1V15JwkAlg;
```
jonasfj, line 69: These need documentation comments. We're not overriding something that already documented.

lib/src/impl_ffi/impl_ffi.digest.dart
```diff
@@ -62,39 +59,100 @@ abstract class _Hash implements Hash {
...
-const Hash sha512 = _Sha512();
+// const HashImpl sha1 = _Sha1();
+// const HashImpl sha256 = _Sha256();
+// const HashImpl sha384 = _Sha384();
+// const HashImpl sha512 = _Sha512();
+
+HashImpl get sha1 => const _Sha1();
+HashImpl get sha256 => const _Sha256();
+HashImpl get sha384 => const _Sha384();
+HashImpl get sha512 => const _Sha512();
```
jonasfj, line 156: What do we need this for?

jonasfj (changes requested):

lib/src/impl_ffi/impl_ffi.digest.dart
HamdaanAliQuatil: Noted 🫡

lib/src/testing/webcrypto/digest.dart
HamdaanAliQuatil: Reverted 🫡

lib/src/impl_js/impl_js.digest.dart
HamdaanAliQuatil: Added `Assertion Error` with the same message

jonasfj (approved):

jonasfj:
Let's ignore the broken windows test and land this.

PR #172, opened 2024-11-08
Update fillRandomBytes

### LINKED ISSUE #105: Avoid @sealed annotation, using `final class` for everything

HamdaanAliQuatil:
Closes #105


test/crypto_subtle_test.dart
```diff
@@ -30,7 +30,7 @@ void main() {
         data.every((e) => e == 0),
         isTrue,
       );
-      fillRandomBytes(data);
+      webCryptImpl.random.fillRandomBytes(data);
```
jonasfj, line 33: This file is supposed to test `lib/src/crypto_subtle.dart`.

Can we not use `subtle.window.crypto.getRandomValues` directly?

HamdaanAliQuatil: imo we should use `subtle.window.crypto.getRandomValues` directly but  `dart:js_interop` does not support `Int64ListToJSInt64Array`. [Reference](https://api.dart.dev/stable/3.2.4/dart-js_interop/)

I've written a workaround for this. The test `Uint64List: not supported type` works as intended.
The test `Uint8List: too long` fails on value `1000000`. The max limit on webcrypto API is `65535` and the error response did not match. [Reference](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues#exceptions). Updated that as well.

test/crypto_subtle_test.dart
```diff
@@ -63,7 +63,7 @@ void main() {
         data.every((e) => e == 0),
         isTrue,
       );
-      subtle.window.crypto.getRandomValues(data.toJS);
+      fillRandomBytes(data);
```
jonasfj, line 66: why are we changing this? `subtle` is still imported, isn't it?

jonasfj (approved):

test/crypto_subtle_test.dart
HamdaanAliQuatil: Taking this to a different PR as discussed 🙌

HamdaanAliQuatil:
@jonasfj all checks pass. we're gtg 🚀

PR #177, opened 2024-11-25
Make OperationError final

jonasfj:


jonasfj:
We could probably use the `@internal` annotation on other members and classes that aren't supposed to be exported, but also aren't private to a library using `_` prefix.

HamdaanAliQuatil (approved):

PR #178, opened 2024-12-02
More support for wasm

koji-1009:
There are several related fixes.

* In the wasm environment, if an invalid value is passed to `getRandomValue`, a JavaScriptError will be thrown. However, JavaScriptError is not public. For this reason, the Error class is catched.
  * ref https://github.com/dart-lang/sdk/issues/55496
  * [JavaScriptError](https://github.com/dart-lang/sdk/blob/3.5.4/sdk/lib/_internal/wasm/lib/errors_patch.dart#L37-L47)
  * [Error class doc](https://api.dart.dev/stable/3.5.4/dart-core/Error-class.html)
* ~Fixed `getRandomValues` argument values are not updated in the wasm environment, so return value is used.~
* Fetch `userAgent` via js_interop. This change can also be implemented by introducing `package:web`.
  * https://github.com/google/webcrypto.dart/pull/180
* (I don't know the cause, this has been happening since Dec 1.) In windows-latest, if I run `flutter pub run test` twice, the first test for firefox fails the second time. To work around this issue and to reduce CI latency, I am separating the coverage jobs.
  * https://github.com/google/webcrypto.dart/pull/179

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -239,7 +239,6 @@ jobs:
           flutter config --no-analytics
       - run: flutter pub get
       - run: flutter pub run webcrypto:setup
-      - run: flutter test
       - run: flutter pub run test -p vm,chrome,firefox -c dart2js,dart2wasm --coverage ./coverage
       # Report collected coverage
       - name: Convert coverage to lcov
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,9 +26,6 @@ jobs:
     name: Linux desktop / Chrome / Firefox
     runs-on: ubuntu-latest
     timeout-minutes: 15
-    strategy:
-      matrix:
-        option: ["", "--wasm"]
     steps:
       - uses: actions/checkout@v4
       - uses: subosito/flutter-action@v2
@@ -43,7 +40,8 @@ jobs:
       - run: flutter pub get
       - run: flutter pub run webcrypto:setup
       - run: flutter test
-      - run: flutter test --platform chrome ${{ matrix.option }}
+      - run: flutter test --platform chrome
+      - run: flutter test --platform chrome --wasm
       - run: xvfb-run flutter test integration_test/webcrypto_test.dart -d linux
         working-directory: ./example
       - uses: nanasess/setup-chromedriver@v2
@@ -59,9 +57,6 @@ jobs:
     name: MacOS 14 desktop / Chrome / Firefox
     runs-on: macos-14 # Test with xcode 15
     timeout-minutes: 15
-    strategy:
-      matrix:
-        option: ["", "--wasm"]
     steps:
       - uses: actions/checkout@v4
       - uses: subosito/flutter-action@v2
@@ -74,7 +69,8 @@ jobs:
       - run: flutter pub get
       - run: flutter pub run webcrypto:setup
       - run: flutter test
-      - run: flutter test --platform chrome ${{ matrix.option }}
+      - run: flutter test --platform chrome
+      - run: flutter test --platform chrome --wasm
       - run: flutter test integration_test/webcrypto_test.dart -d macos
         working-directory: ./example
       # TODO: Enable chromedriver testing on MacOS when it works reliably
@@ -129,9 +125,6 @@ jobs:
     name: Windows desktop / Chrome / Firefox
     runs-on: windows-latest
     timeout-minutes: 15
-    # strategy:
-    #   matrix:
-    #     option: [ "", "--wasm"]
     steps:
       - uses: ilammy/setup-nasm@v1
       - uses: actions/checkout@v4
@@ -145,7 +138,8 @@ jobs:
       - run: flutter pub get
       - run: flutter pub run webcrypto:setup
       - run: flutter test
-      #- run: flutter test --platform chrome ${{ matrix.option }}
+      #- run: flutter test --platform chrome
+      #- run: flutter test --platform chrome --wasm
       - run: flutter test integration_test/webcrypto_test.dart -d windows
         working-directory: ./example
       - uses: nanasess/setup-chromedriver@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -239,7 +239,7 @@ jobs:
           flutter config --no-analytics
       - run: flutter pub get
       - run: flutter pub run webcrypto:setup
-      - run: flutter pub run test -p vm,chrome,firefox -c dart2js,dart2wasm --coverage ./coverage
+      - run: flutter pub run test -p vm,chrome,firefox --coverage ./coverage
       # Report collected coverage
       - name: Convert coverage to lcov
         run: dart run coverage:format_coverage -i ./coverage -o ./coverage/lcov.info --lcov --report-on lib/
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -94,9 +94,6 @@ jobs:
     name: MacOS 15 desktop / Chrome / Firefox
     runs-on: macos-15 # Test with xcode 16
     timeout-minutes: 15
-    strategy:
-      matrix:
-        option: ["", "--wasm"]
     steps:
       - uses: actions/checkout@v4
       - uses: subosito/flutter-action@v2
@@ -109,7 +106,8 @@ jobs:
       - run: flutter pub get
       - run: flutter pub run webcrypto:setup
       - run: flutter test
-      - run: flutter test --platform chrome ${{ matrix.option }}
+      - run: flutter test --platform chrome
+      - run: flutter test --platform chrome --wasm
       - run: flutter test integration_test/webcrypto_test.dart -d macos
         working-directory: ./example
       # TODO: Enable chromedriver testing on MacOS when it works reliably
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
   lints:
     name: Format + Lints
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 15
     steps:
       - uses: actions/checkout@v4
       - uses: subosito/flutter-action@v2
@@ -25,7 +25,7 @@ jobs:
   linux:
     name: Linux desktop / Chrome / Firefox
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 15
     strategy:
       matrix:
         option: ["", "--wasm"]
@@ -58,7 +58,7 @@ jobs:
   macos-14:
     name: MacOS 14 desktop / Chrome / Firefox
     runs-on: macos-14 # Test with xcode 15
-    timeout-minutes: 10
+    timeout-minutes: 15
     strategy:
       matrix:
         option: ["", "--wasm"]
@@ -97,7 +97,7 @@ jobs:
   macos-15:
     name: MacOS 15 desktop / Chrome / Firefox
     runs-on: macos-15 # Test with xcode 16
-    timeout-minutes: 10
+    timeout-minutes: 15
     strategy:
       matrix:
         option: ["", "--wasm"]
@@ -128,7 +128,7 @@ jobs:
   windows:
     name: Windows desktop / Chrome / Firefox
     runs-on: windows-latest
-    timeout-minutes: 10
+    timeout-minutes: 15
     # strategy:
     #   matrix:
     #     option: [ "", "--wasm"]
@@ -161,7 +161,7 @@ jobs:
   ios:
     name: iOS emulator (iPhone)
     runs-on: macos-14
-    timeout-minutes: 10
+    timeout-minutes: 15
     steps:
       - uses: actions/checkout@v4
       - uses: subosito/flutter-action@v2
@@ -180,7 +180,7 @@ jobs:
   android:
     name: Android emulator
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 15
     steps:
       - uses: actions/checkout@v4
       - uses: subosito/flutter-action@v2
@@ -209,7 +209,7 @@ jobs:
   linux-coverage:
     name: Linux desktop / Chrome / Firefox (coverage)
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 15
     steps:
       - uses: actions/checkout@v4
       - uses: subosito/flutter-action@v2
@@ -223,7 +223,6 @@ jobs:
           flutter config --no-analytics
       - run: flutter pub get
       - run: flutter pub run webcrypto:setup
-      - uses: nanasess/setup-chromedriver@v2
       - run: xvfb-run flutter pub run test -p vm,chrome,firefox --coverage ./coverage
       # Report collected coverage
       - name: Convert coverage to lcov
@@ -235,7 +234,7 @@ jobs:
   windows-coverage:
     name: Windows desktop / Chrome / Firefox (coverage)
     runs-on: windows-latest
-    timeout-minutes: 10
+    timeout-minutes: 15
     steps:
       - uses: ilammy/setup-nasm@v1
       - uses: actions/checkout@v4
@@ -260,7 +259,7 @@ jobs:
   coverage:
     name: Report to coveralls
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 15
     needs:
       - linux-coverage
       - windows-coverage
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,6 +12,7 @@ jobs:
   lints:
     name: Format + Lints
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


lib/src/crypto_subtle.dart
```diff
@@ -334,23 +334,23 @@ extension type JSRsaOtherPrimesInfo(JSObject _) implements JSObject {
 
 TypedData getRandomValues(TypedData array) {
   if (array is Uint8List) {
-    window.crypto.getRandomValues(array.toJS);
-    return array;
+    final values = window.crypto.getRandomValues(array.toJS);
+    return (values as JSUint8Array).toDart;
```
jonasfj, line 338: Is there any need for this cast, doesn't `getRandomValues` always return the first argument that it is given?

lib/src/impl_js/impl_js.random.dart
```diff
@@ -16,8 +16,34 @@ part of 'impl_js.dart';
 
 void fillRandomBytes(TypedData destination) {
   try {
-    subtle.getRandomValues(destination);
+    if (destination is Uint8List) {
+      final newValues = subtle.getRandomValues(destination);
+      destination.setAll(0, newValues as Uint8List);
```
jonasfj, line 21: What is this good for? `getRandomValues` writes to the first argument it is given, does it not?

test/crypto_subtle_test.dart
```diff
@@ -261,7 +275,7 @@ void main() {
               false,
             )
             .having(
-              (key) => key.privateKey.usages,
+              (key) => key.privateKey.usages.toDartList,
```
jonasfj, line 278: Can we do the `.usages.toDartList` as a separate PR.

It seems to me like we can get this passing all existing tests. That should make it easy to land, and it'll reduce the size of this PR.

lib/src/impl_js/impl_js.random.dart
```diff
@@ -16,8 +16,34 @@ part of 'impl_js.dart';
...
+      throw UnsupportedError(
+        'Unsupported TypedData type: ${destination.runtimeType}',
+      );
+    }
   } on subtle.JSDomException catch (e) {
     throw _translateDomException(e);
+  } on UnsupportedError {
+    rethrow;
+  } on Error catch (e) {
+    throw _translateJavaScriptException(e);
```
jonasfj, line 47: I'm not necessarily convinced this is a good idea.

We don't want to catch errors that we're not supposed to catch.

Hmm, I'm not sure exactly how we should handle this.

I'm guessing it means that the `_handleDomException` logic we have also won't work as intended.

I wonder if it's:
(a) Possible to get access to the JSObject that was thrown (somehow).
(b) Is there a bug for how error handling differs between js and wasm.

I'm betting that we don't have good enough tests in this package for all the exception cases, hence, also why a lot of tests didn't start failing when `_handleDomException` doesn't work as expected anymore.

Maybe, we need to put the error handling logic that differs into `crypto_suble.dart` such that `impl_js/` doesn't have to care whether it's running in dart2js or dart2wasm.

But I'm surprised that error handling differs between js and wasm, maybe we're doing something wrong :rofl:

lib/src/testing/utils/detected_runtime_js_interop.dart
```diff
@@ -12,8 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// ignore: avoid_web_libraries_in_flutter
-import 'dart:html' show window;
+import 'dart:js_interop';
+
+@JS('window.navigator.userAgent')
+external String get userAgent;
```
jonasfj, line 18: Can we do this in a separate PR, I'm guessing that this change won't break existing tests and it would be easy to land in a separate PR -- thus, making this PR smaller.

jonasfj (commented):
:rocket: Thanks for diving into this, I don't think neither me or @HamdaanAliQuatil has had much time to play with wasm yet.

I'm not sure exactly how to get the error handling, bit right, so let's explore it a bit -- see if there are any related bugs on the issue, etc.

My other suggestion would be to try and split as many of changes into tiny PRs. Those are much easier to land, and it'll make easier for us to focus on the complex issue in this PR.

.github/workflows/test.yml
```diff
@@ -207,12 +206,64 @@ jobs:
           api-level: 28
           working-directory: ./example
           script: flutter test integration_test/webcrypto_test.dart -d android
+  linux-coverage:
```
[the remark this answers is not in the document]
koji-1009: I made a separate PR.
https://github.com/google/webcrypto.dart/pull/179

lib/src/impl_js/impl_js.random.dart
koji-1009: It seems that dart2js overwrites the value, but dart2wasm does not. Could you please see that the `getRandomValues` test fails once as I will submit a PR later that runs with the `--wasm` option in ci?

test/crypto_subtle_test.dart
koji-1009: I made a separate PR.
https://github.com/google/webcrypto.dart/pull/181

lib/src/testing/utils/detected_runtime_js_interop.dart
koji-1009: I made a separate PR.
https://github.com/google/webcrypto.dart/pull/180

lib/src/impl_js/impl_js.random.dart
jonasfj: doesn't that hint at a more fundamental issue.

If you read the documentation for `getRandomValues` it's very clear that it fills the first argument with random bytes:

https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues

It even says it's updated in-place:
image


----------

If that's not working when compiling to wasm, then what is happening?

I'd suggest trying to make a minimal example. Because this shouldn't be happening, if we can't pass typedarrays correctly to JS from inside wasm, then something is very wrong.

lib/src/impl_js/impl_js.random.dart
jonasfj: Thanks for referencing https://github.com/dart-lang/sdk/issues/55496

I'd suggest we look at `toString()`, because the only error we want to handle here is one that is `_JavaScriptError`. We might even want to make this code not run when it's compiled with dart2js (not sure if we can detect wasm vs js, though :rofl:)

koji-1009: I would modify the code as follows 👍 
I wasn't expecting `Error` handling, so I was confused when implementing it. Thanks for the review.

```dart
} on Error catch (e) {
    final errorName = e.toString();
    if (errorName != 'JavaScriptError') {
        rethrow;
    }

    throw _translateJavaScriptException(e);
}
```

jonasfj: I think this is a good idea, we probably have to do that in `_handleDomException` and in `getRandomValues` inside `lib/src/crypto_subtle.dart`.

lib/src/impl_js/impl_js.random.dart
koji-1009: I tried it and the following code works.
In the case of dart2js, the `array.toJS` reflected the value in the `array`, but not in the case of dart2wasm.

```dart
final values = array.toJS;
window.crypto.getRandomValues(values);
return values.toDart;
```

jonasfj: Can we make a gist that contains the smallest possible test case where:
```
$ dart test -p chrome -c dart2wasm
```
and
```
$ dart test -p chrome -c dart2js
```
behaves differently.

Then let's report a bug against the SDK and see if we can find a workaround.

--------

If there really is a thing where modifications to a `Uint8List` passed to Javascript aren't reflected back in Dart that's pretty bad. And we'd need to review the entire code base for all places where we pass something to Javascript to check that we don't rely on side effects.

lib/src/impl_js/impl_js.utils.dart
```diff
@@ -99,6 +99,13 @@ Object _translateDomException(
       '"${e.name}", message: $message');
 }
 
+Object _translateJavaScriptException(Error e) {
```
jonasfj, line 102: Let's give this a documentation comment.

Let's link to the issue in the SDK, that way when we discover that the issue is resolved, we can work to fix this hack.

lib/src/impl_js/impl_js.utils.dart
```diff
@@ -99,6 +99,13 @@ Object _translateDomException(
...
+  // dart2wasm throws _JavaScriptError, but _JavaScriptError is not exposed.
+  final errorMessage = e.toString();
+  final message = 'browser threw "$errorMessage"';
+  return ArgumentError(message);
```
jonasfj, line 106: I don't know what we should throw here. This could be `UnsupportedError`, `ArgumentError` or even `OperationError`.

We might need to throw a new `UnknownError` that contains a message explaining how this version of `package:webcrypto` cannot distinguish between the different kinds of errors thrown by Javascript.

Possibly with a link to an issue in our issue tracker on this repository.

lib/src/impl_js/impl_js.random.dart
koji-1009: https://github.com/dart-lang/sdk/issues/59651

koji-1009: It was shared that this behavior is to be expected as per the [dart doc](https://api.dart.dev/dart-js_interop/ListToJSArray/toJS.html) and [article's warning](https://dart.dev/interop/js-interop/js-types). For this reason, it would be a good idea to add a process to overwrite the list on the dart side.

jonasfj: Ah, I didn't know that `.toJS` could return a clone.

We should probably review all places in the code where we do a `.toJS` call.

jonasfj: hmm, I guess we could work around this as follows:
```dart
import 'dart:js_interop';
import 'dart:typed_data';

import 'package:web/web.dart' as web;
import 'package:test/test.dart';

void main() {
  test('calculate', () {
    final array = Uint8List(16);
    expect(
      array.any((e) => e == 0),
      isTrue,
    );

    final jsArray = array.toJS;
    web.window.crypto.getRandomValues(jsArray);

    // .toJS may cause a clone, so we can't know if
    // jsArray.toDart is the same as the original array.
    final dartArray = jsArray.toDart;
    if (dartArray == array) {
      print('We do not need to do anything here');
    } else {
      array.setAll(0, dartArray);
    }

    expect(
      array.any((e) => e != 0),
      isTrue,
    );
  });
}
```

probably we can pick some better variable names. But there is no need to always do `.setAll`.

lib/src/impl_js/impl_js.utils.dart
```diff
@@ -99,6 +99,29 @@ Object _translateDomException(
...
 }
 
+/// Convert [Error] to [UnknownError].
+/// dart2wasm throws _JavaScriptError, but _JavaScriptError is not exposed.
+///
+/// [1]: https://github.com/dart-lang/sdk/issues/55496
+/// [2]: https://api.dart.dev/stable/latest/dart-core/Error-class.html
+Object _translateJavaScriptException(Error e) {
+  final errorMessage = e.toString();
+  final message = 'browser threw "$errorMessage"';
```
jonasfj, line 109: Is `errorMessage` not always going to be `"JavaScriptError"` ?

koji-1009: The UnknownError message can also be a constant, so it has been fixed. [4e8e9de](https://github.com/google/webcrypto.dart/pull/178/commits/4e8e9de977dd1b6dddde98c20c526c3d59623ce1)

lib/src/impl_js/impl_js.random.dart
koji-1009: Oops, calling `import 'package:flutter/foundation.dart' show kIsWasm;` on `crypto_subtle.dart` causes an error.

lib/src/crypto_subtle.dart
```diff
@@ -333,31 +333,49 @@ extension type JSRsaOtherPrimesInfo(JSObject _) implements JSObject {
 }
 
 TypedData getRandomValues(TypedData array) {
+  // Since dart2wasm does not reflect values in the array, use setAll to reflect them.
+  // see: https://github.com/dart-lang/sdk/issues/59651
```
jonasfj, line 337:
```suggestion
  // The `.toJS` on `Uint8List` (and friends) may:
  //   * cast,
  //   * create a wrapper, or,
  //   * clone.
  // See: https://api.dart.dev/dart-js_interop/Uint8ListToJSUint8Array/toJS.html
  //
  // Thus, when we do `.toJS` and pass the resulting object to `getRandomValues`,
  // we don't know if `.toJS` simply cast or created a wrapper such that changes
  // made by `getRandomBytes` are reflected in `array`.
  //
  // For this reason, we must use `.setAll` to copy the values into `array`, if it
  // was not cast.
  // 
  // See also: https://github.com/dart-lang/sdk/issues/59651
```

lib/src/crypto_subtle.dart
```diff
@@ -333,31 +333,49 @@ extension type JSRsaOtherPrimesInfo(JSObject _) implements JSObject {
...
   if (array is Uint8List) {
-    window.crypto.getRandomValues(array.toJS);
-    return array;
+    final values = array.toJS;
+    window.crypto.getRandomValues(values);
+    if (array != values.toDart) {
+      array.setAll(0, values.toDart);
+    }
```
jonasfj, line 343: Let's avoid calling `.toDart` twice, in wasm mode it'll create a wrapper, there is no need to create the wrapper twice.
```suggestion
    final jsarray = array.toJS;
    window.crypto.getRandomValues(jsarray);
    final dartArray = jsarray.toDart;
    if (array != dartArray) {
      array.setAll(0, dartArray);
    }
```

See: https://api.dart.dev/dart-js_interop/JSUint8ArrayToUint8List/toDart.html

lib/src/impl_js/impl_js.utils.dart
```diff
@@ -99,6 +99,25 @@ Object _translateDomException(
...
+///
+/// [1]: https://github.com/dart-lang/sdk/issues/55496
+/// [2]: https://api.dart.dev/stable/latest/dart-core/Error-class.html
+Object _translateJavaScriptException() {
+  return UnknownError();
+}
+
+/// Error class for handling JavaScriptError that occurred in package:webcrypto.
+class UnknownError extends Error {
+  UnknownError();
```
jonasfj, line 113: Let's not have anyone else creating instances of this error.
```suggestion
  UnknownError._();
```

test/crypto_subtle_test.dart
```diff
@@ -68,32 +73,60 @@ void main() {
         data.any((e) => e != 0),
         isTrue,
       );
-    });
+    }, skip: 'dart2wasm');
```
jonasfj, line 76: I'm curious, why do we need to skip this?

lib/src/impl_js/impl_js.random.dart
jonasfj: We can just use `bool.fromEnvironment('dart.tool.dart2wasm')` it's the same :rofl:

lib/src/crypto_subtle.dart
```diff
@@ -333,31 +333,49 @@ extension type JSRsaOtherPrimesInfo(JSObject _) implements JSObject {
...
   } else {
-    throw ArgumentError.value(
-      array,
-      'array',
-      'Unsupported TypedData type',
-    );
+    throw UnsupportedError('Unsupported TypedData type ${array.runtimeType}');
```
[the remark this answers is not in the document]
koji-1009: I noticed that when generating unsupported lists (e.g. `Uint64List`) in dart2js, an UnsupportedError is thrown. For example, the following test succeeds.

```dart
test('Uint64List: not supported type', () {
  expect(
    () => Uint64List(32),
    throwsA(
      isA<UnsupportedError>(),
    ),
  );
});
```

The test fails in dart2wasm.

---

There is no JS class corresponding to `Uint64List`. Therefore, when `fillRandomBytes(Uint64List(32))` is called in dart2wasm, the process reaches this code.

I changed it to UnsupportedError to keep the process consistent between dart2js and dart2wasm. It is possible to split a PR.

test/crypto_subtle_test.dart
koji-1009: This is because with dart2wasm, the value is not reflected in the UInt8List. All `data` values are 0.

lib/src/crypto_subtle.dart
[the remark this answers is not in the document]
koji-1009: I added test case. [4efab3b](https://github.com/google/webcrypto.dart/pull/178/commits/4efab3bd9d42e65d2b554efc6d5025348075c027)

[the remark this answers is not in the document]
koji-1009: And restore previous code. The addition of the test makes it easier to understand the scope of the impact. [bd9425a](https://github.com/google/webcrypto.dart/pull/178/commits/bd9425a5b015565627dfc08f16618d659e4312c9)

PR #179, opened 2024-12-02
chore: Split ci coverage job

koji-1009:
ref #178 

* Separation of jobs to check coverage
* Setting timeout-minutes
  * Set to exit when flutter run test fails

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,6 +12,7 @@ jobs:
   lints:
     name: Format + Lints
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@v4
       - uses: subosito/flutter-action@v2
@@ -24,6 +25,7 @@ jobs:
   linux:
     name: Linux desktop / Chrome / Firefox
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@v4
       - uses: subosito/flutter-action@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -12,6 +12,7 @@ jobs:
   lints:
     name: Format + Lints
     runs-on: ubuntu-latest
+    timeout-minutes: 10
```
jonasfj, line 15: Looking at the last successful run on master, one of the jobs did take 11mins:
https://github.com/google/webcrypto.dart/actions/runs/12068757318

So maybe we should set the timeouts to 15mins.

Though I agree, it would be nice if we could engineer the jobs to finish faster.
Then again, we don't want to have too many jobs, it can be a resource waste and even slow because we'd have to wait for the jobs to start.

.github/workflows/test.yml
```diff
@@ -207,12 +194,64 @@ jobs:
...
+          channel: 'stable'
+          cache: true
+      - name: Configure Flutter
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y ninja-build libgtk-3-dev
+          flutter config --no-analytics
+      - run: flutter pub get
+      - run: flutter pub run webcrypto:setup
+      - uses: nanasess/setup-chromedriver@v2
```
jonasfj, line 214: We shouldn't need chromedriver for this
```suggestion
```

jonasfj (commented):
Let's bump the timeouts to 15 min

.github/workflows/test.yml
koji-1009: [bfe006b](https://github.com/google/webcrypto.dart/pull/179/commits/bfe006bd2238b97d2aa24bad3c550ca3374296dd) 

Updated at 15 minutes.
Splitting the job will make it shorter than 10 minutes, but it is certainly possible to go over due to the effects of increasing the number of tests, etc. I think 15 minutes is reasonable. Thanks.

jonasfj (approved):
r+

PR #181, opened 2024-12-02
test: Modified to a more strict type

koji-1009:
ref https://github.com/google/webcrypto.dart/pull/178

Support for more strict type requirements in wasm build.

### DIFF
--- a/test/crypto_subtle_test.dart
+++ b/test/crypto_subtle_test.dart
@@ -122,7 +122,7 @@ void main() {
               false,
             )
             .having(
-              (key) => key.usages,
+              (key) => key.usages.toDartList,
               'usages',
               containsAll(['encrypt', 'decrypt']),
             ),
@@ -198,7 +198,7 @@ void main() {
               true,
             )
             .having(
-              (key) => key.publicKey.usages,
+              (key) => key.publicKey.usages.toDartList,
               'publicKey.usages',
               ['encrypt'],
             )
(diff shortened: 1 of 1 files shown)


jonasfj (approved):

jonasfj:
Is this a known issue? If not it might be interesting to make a minimum reproducible example and file an issue against the SDK.

It seems like it would be nice if there was a page somewhere with known compatibility issues between Dart when compiled the JS and wasm, respectively.

koji-1009:
Comparing `JSArray<JSString>` with `<String>[]` shows that it does not implements `Iterable`. Running the test on the `feat/replace_html` branch will result in the following test fail.

```
To run this test again: /Users/koji/flutter/bin/cache/dart-sdk/bin/dart test test/crypto_subtle_test.dart -p chrome --plain-name 'crypto getRandomValues: not supported type'
00:09 +1415 -14: test/crypto_subtle_test.dart: crypto.subtle generateCryptoKey: success [E]
  Expected: <Instance of 'JSValue'> with `type`: 'secret' and `extractable`: <false> and `usages`: contains all of ['encrypt', 'decrypt']
    Actual: JSValue:<[object CryptoKey]>
     Which: has `usages` with value JSValue:<encrypt,decrypt> which not an <Instance of 'Iterable'>
```

I think it is because `JSString` and `String` are converted by js_interop, but `JSArray` and `List` are not. it is true that dart2wasm is more type strict than dart2js, but it is certainly not convenient.

PR #195, opened 2025-02-28
Replace KeyPair class with record `typedef KeyPair<T,S>`

### LINKED ISSUE #191: Consider removing `KeyPair` class in favor of a record `typedef KeyPair<T,S> = ({T privateKey, S publicKey});`
We could do:
```dart
typedef KeyPair<T,S> = ({T privateKey, S publicKey});
```

Instead of having a `KeyPair` class, it's not like we need the class for anything, all it does is create a type that doesn't have to exist.

It would be a cleaner API t use a record.

Given that there is no public constructor for `KeyPair` and the fact that it is `final`; I'm not even sure it's really a breaking change.

---------

Making this change is probably easy. But I think we should land it along with a breaking version bump, so `0.5.x -> 0.6.0` or something like that.

HamdaanAliQuatil:
Closes #191

### DIFF
--- a/lib/src/testing/utils/testrunner.dart
+++ b/lib/src/testing/utils/testrunner.dart
@@ -213,15 +213,7 @@ typedef DeriveBitsFn<T> = Future<List<int>> Function(
   Map<String, dynamic> deriveParams,
 );
 
-class _KeyPair<S, T> implements KeyPair<S, T> {
-  @override
-  final S privateKey;
-
-  @override
-  final T publicKey;
-
-  _KeyPair({required this.privateKey, required this.publicKey});
-}
+typedef KeyPair<S, T> = ({S privateKey, T publicKey});
 
 @sealed
 class TestRunner<PrivateKey, PublicKey> {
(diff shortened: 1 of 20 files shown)


jonasfj:
This is technically breaking right, but not very breaking.

I was contemplating if we should group with other breaking changes. But maybe we should just decide that it's not that breaking.

jonasfj (commented):
Let's update the `CHANGELOG.md` so that we have entry for this change, ready for the next release.

This is technically breaking, but probably so little breaking that it won't affect anyone.

lib/src/testing/webcrypto/ecdh.dart
```diff
@@ -17,16 +17,7 @@ import '../utils/utils.dart';
...
-    implements KeyPair<S, T> {
-  @override
-  final S privateKey;
-
-  @override
-  final T publicKey;
-
-  _KeyPair({required this.privateKey, required this.publicKey});
-}
+typedef KeyPair<S, T> = ({S privateKey, T publicKey});
```
jonasfj, line 20: where is this used?

lib/src/testing/utils/testrunner.dart
```diff
@@ -213,15 +213,7 @@ typedef DeriveBitsFn<T> = Future<List<int>> Function(
...
-class _KeyPair<S, T> implements KeyPair<S, T> {
-  @override
-  final S privateKey;
-
-  @override
-  final T publicKey;
-
-  _KeyPair({required this.privateKey, required this.publicKey});
-}
+typedef KeyPair<S, T> = ({S privateKey, T publicKey});
```
jonasfj, line 216: where is this used?

jonasfj (approved):

jonasfj (approved):

PR #197, opened 2025-03-10
Adding support for randomUUID

### LINKED ISSUE #187: Consider adding support for randomUUID
See: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

And https://w3c.github.io/webcrypto/

We also need to explore browser support.

Dhruv-Maradiya:
### Changes
- Added `randomUUID()` method.
- Followed the [W3C Web Crypto API](https://w3c.github.io/webcrypto/#Crypto-method-randomUUID) specification.
- Checked BoringSSL but couldn't find an existing `randomUUID` implementation, so implemented it based on the W3C spec.
- Verified browser compatibility using [Can I use](https://caniuse.com/?search=randomUUID)—support looks good.

### Testing
- Added test cases to validate UUID generation.
- Currently, the test runs a loop 100 times to generate UUIDs—open to feedback on whether this is the best approach.

### Notes
- Fixes #187  
- I'm still exploring the repository, so let me know if I missed anything.

### DIFF
--- a/lib/src/impl_ffi/impl_ffi.random.dart
+++ b/lib/src/impl_ffi/impl_ffi.random.dart
@@ -30,4 +30,28 @@ final class _RandomImpl implements RandomImpl {
       dest.setAll(0, out.asTypedList(dest.length));
     });
   }
+
+  @override
+  String randomUUID() {
+    return _Scope.sync((scope) {
+      final out = scope<ffi.Uint8>(16);
+      _checkOp(ssl.RAND_bytes(out, 16) == 1);
+
+      var bytes = out.asTypedList(16);
+
+      // Set the 4 most significant bits of bytes[6] to 0100.
+      bytes[6] = (bytes[6] & 0x0F) | 0x40;
+
+      // Set the 2 most significant bits of bytes[8] to 10.
+      bytes[8] = (bytes[8] & 0x3F) | 0x80;
+
+      return bytes
+          .map((e) => e.toRadixString(16).padLeft(2, '0'))
+          .join()
+          .replaceAllMapped(
+            RegExp(r'(\w{8})(\w{4})(\w{4})(\w{4})(\w{12})'),
+            (m) => '${m[1]}-${m[2]}-${m[3]}-${m[4]}-${m[5]}',
+          );
+    });
+  }
 }
--- a/lib/src/testing/webcrypto/random.dart
+++ b/lib/src/testing/webcrypto/random.dart
@@ -24,6 +24,14 @@ void isNotAllZero(TypedData data) {
   check(data.buffer.asUint8List().any((b) => b != 0));
 }
 
+void isValidUUID(String uuid) {
+  check(uuid.length == 36);
+  check(uuid[8] == '-');
+  check(uuid[13] == '-');
+  check(uuid[18] == '-');
+  check(uuid[23] == '-');
+}
+
 void main() => tests().runTests();
 
 /// Tests, exported for use in `../run_all_tests.dart`.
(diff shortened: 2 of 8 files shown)


lib/src/impl_ffi/impl_ffi.random.dart
```diff
@@ -30,4 +30,28 @@ final class _RandomImpl implements RandomImpl {
...
+      // Set the 2 most significant bits of bytes[8] to 10.
+      bytes[8] = (bytes[8] & 0x3F) | 0x80;
+
+      return bytes
+          .map((e) => e.toRadixString(16).padLeft(2, '0'))
+          .join()
+          .replaceAllMapped(
+            RegExp(r'(\w{8})(\w{4})(\w{4})(\w{4})(\w{12})'),
+            (m) => '${m[1]}-${m[2]}-${m[3]}-${m[4]}-${m[5]}',
+          );
```
jonasfj, line 54: I've done this other places the classical implementation is:
```
final bytes = _randomBytes(16);
  // Set V4 bits according to:
  // https://tools.ietf.org/html/rfc4122#section-4.4
  bytes[6] = (bytes[6] & 0x0f) | 0x40;
  bytes[8] = (bytes[8] & 0x3f) | 0x80;

  // Encode as UUIDv4
  final s = hex.encode(bytes).substring;
  return '${s(0, 8)}-${s(8, 12)}-${s(12, 16)}-${s(16, 20)}-${s(20)}';
```

Maybe, using `toRadixString` is nicer, since it avoids a dependency on `package:convert`.

lib/src/impl_ffi/impl_ffi.random.dart
```diff
@@ -30,4 +30,28 @@ final class _RandomImpl implements RandomImpl {
...
     });
   }
+
+  @override
+  String randomUUID() {
+    return _Scope.sync((scope) {
+      final out = scope<ffi.Uint8>(16);
+      _checkOp(ssl.RAND_bytes(out, 16) == 1);
+
+      var bytes = out.asTypedList(16);
```
jonasfj, line 40: Maybe, we could just use `fillRandomBytes`, then we have less FFI stuff going on.

lib/src/testing/webcrypto/random.dart
```diff
@@ -24,6 +24,14 @@ void isNotAllZero(TypedData data) {
   check(data.buffer.asUint8List().any((b) => b != 0));
 }
 
+void isValidUUID(String uuid) {
+  check(uuid.length == 36);
+  check(uuid[8] == '-');
+  check(uuid[13] == '-');
+  check(uuid[18] == '-');
+  check(uuid[23] == '-');
```
jonasfj, line 32: Check the bits too

lib/src/webcrypto/webcrypto.random.dart
```diff
@@ -48,3 +48,18 @@ void fillRandomBytes(
...
+/// import 'package:webcrypto/webcrypto.dart';
+///
+/// // Generate a random UUID.
+/// void main() {
+///   final uuid = randomUUID();
+///   print(uuid);
+/// }
+///
+/// ```
+String randomUUID() => webCryptImpl.random.randomUUID();
```
jonasfj, line 65: This generates a UUIDv4, there is a specification let's explain that:
https://tools.ietf.org/html/rfc4122#section-4.4

jonasfj, line 65: The name of the function in Dart should follow conventions:
https://dart.dev/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words

`randomUuid()`, however, ugly :see_no_evil: 

We could choose to just call it `uuid()`, I don't know. We do change names from Javascript when implementing them in Dart, like how `getRandomValues` became `fillRandomBytes`.

jonasfj (changes requested):

lib/src/webcrypto/webcrypto.random.dart
jonasfj: @HamdaanAliQuatil any thoughts on the name?

HamdaanAliQuatil: `uuid` looks good.

To add to this - based on the last discussion on `randomUUID` on `w3c`, support for UUIDv7 (or other variants), if added, would be in the form of parameters.

If someday webcrypto introduces ULID, it will likely be another in another class and that would be an issue for another day.

As for now, and for the near future as well, `uuid` solves our purpose.

mosuem:
Closing in favor of #290

Issue #198, opened 2025-05-05
Optimize performance of resource-intensive cryptographic operations using Isolates


HamdaanAliQuatil:
Several cryptographic operations are computationally expensive and block the main thread, causing UI Janks and a poor user experience.

The goal is to identify such operations and isolate the intensive computation to background threads.
Use [dart:isolates](https://dart.dev/language/isolates) to create cheap isolates.

jonasfj:
We need to determine what is worth moving off the main thread and what isn't.

I think a clear example of something definitely worth moving is RSA key generation.

* Not much data to copy between isolates
* Finding large primes is expensive!

I clear example of something not worth moving is AES key generation.

* Generation of random data
* Only a few bytes at most
* Dirt cheap, cost of starting an isolate and communicating between isolates greatly exceeds generating a few random bytes.

Unless, of course the source of secure random bytes is empty and the thread is blocked while reading from `/dev/srandom`. But that's kind of an extreme corner case -- I think once BoringSSL is initialized with an initial seed it has an infinite source of random data. So let's not optimize for extreme cases like this, where moving the mouse generates the randomness that wasn't there.

------

I don't know the exact criteria for moving an operation off the main thread. So let's begin that discussion here.

One thing worth keeping in mind is that even filling a large array with random data, or computing the hash for a large blob, is probably extremely cheap. And we could do it in 4MB increments with `await Future.delayed(Duration(0))` in between each increment. Ensuring that there is time for other operations to run.

That doesn't offload the operation to another isolate. It merely splits large operations into multiple microtasks running on the same isolate. But by splitting into separate microtasks we avoid blocking UI rendering, etc.

So we have 3 options for an operation:
 * (a) Do it on the main thread synchronously (what we currently do).
 * (b) Do it on the main thread asynchronously, by doing it in small increments interleaved with other microtasks.
 * (c) Do it on an background isolate, copying the data to said isolate.

While (c) is probably least blocking, it also carries the most overhead. Imagine processing a stream, we'd have to copy event block we get from the stream to the background isolate. Or at least pass an FFI pointer to the block to the isolate. That's a lot of isolate messages going back and forth. Because we'd also have to support a reasonable level of bacl-pressure on the stream, hence, we'd also need to send ack messages back for every block.

[...]

PR #201, opened 2025-05-29
Move EC Key Generation off the main-thread with Isolates

HamdaanAliQuatil:
Part of https://github.com/google/webcrypto.dart/issues/198

EC key generation is faster and less likely to block the main thread because it uses scalar multiplication on elliptic curves, which is significantly less expensive than the prime generation and modular arithmetic involved in RSA. We're making a similar change here to ensure consistency across key generation paths and to ensure responsiveness.

### DIFF
--- a/lib/src/impl_ffi/impl_ffi.dart
+++ b/lib/src/impl_ffi/impl_ffi.dart
@@ -20,6 +20,7 @@ import 'dart:async';
 import 'dart:ffi' show Allocator;
 import 'dart:typed_data';
 import 'dart:convert' show utf8, base64Url;
+import 'dart:isolate';
 import 'dart:ffi' as ffi;
 import 'dart:math' as math;
 import 'package:meta/meta.dart';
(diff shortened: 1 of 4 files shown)


jonasfj:
It's been ages since looked at EC stuff, but as I recall creating private key is dirt cheap, the most expensive part is deriving the public key... Thus, if the computational effort is not import public key, because you usually only store a compressed format... I don't recall the details.

Anyways, consistency shouldn't determine if offload to a background isolate or not. Computational cost should.

If isolate overhead exceeds the cost the offloading is pointless, sha256 is one such case. Copying the data is probably more expensive, than computing on main thread. Or at least close to the same cost.

Where RSA key generation is expensive, compared copying operations required to move off the main thread.

I don't know a good metric for whether something is worth moving is. But this probably what we should start by figuring out.

I propose we continue that discussion in 
https://github.com/google/webcrypto.dart/issues/198

PR #204, opened 2025-06-28
Add Script to Update BoringSSL

HamdaanAliQuatil:


.github/workflows/update-boringssl.yml
```diff
@@ -0,0 +1,222 @@
...
+    
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
```
jonasfj, line 29: We should consider rewriting the script to Dart or bash.

.github/workflows/update-boringssl.yml
```diff
@@ -0,0 +1,222 @@
...
+            LATEST_REVISION="${{ github.event.inputs.boringssl_revision }}"
+            echo "Using manually specified revision: $LATEST_REVISION"
+          else
+            LATEST_REVISION=$(git rev-parse HEAD)
+            echo "Latest BoringSSL revision: $LATEST_REVISION"
+          fi
+          
+          # Get current revision from the Python script
+          CURRENT_REVISION=$(grep "BORINGSSL_REVISION = " tool/update-boringssl.py | cut -d"'" -f2)
+          echo "Current revision in script: $CURRENT_REVISION"
```
jonasfj, line 61: Why don't we put this into a bash script in `tool/`,

If we want we can just put the current revision in `lib/src/third_party/boringssl/REVISION` or something like that, there is no need for it to be hardcoded in the python script.

.github/workflows/update-boringssl.yml
```diff
@@ -0,0 +1,222 @@
...
+
+      - name: Update BoringSSL revision in script
+        if: steps.get-revision.outputs.needs_update == 'true'
+        run: |
+          # Update the BORINGSSL_REVISION in the Python script
+          sed -i "s/BORINGSSL_REVISION = '[^']*'/BORINGSSL_REVISION = '${{ steps.get-revision.outputs.latest_revision }}'/" tool/update-boringssl.py
+          
+          # Verify the change
+          echo "Updated revision in script:"
+          grep "BORINGSSL_REVISION = " tool/update-boringssl.py
```
jonasfj, line 86: Again why not make a bash script in `tool/bump-boringssl-revision.sh`

Or if we just store the revision in a `REVISION` file, then we don't need a script :D

.github/workflows/update-boringssl.yml
```diff
@@ -0,0 +1,222 @@
...
+
+      - name: Run BoringSSL update script
+        if: steps.get-revision.outputs.needs_update == 'true'
+        run: |
+          # Step 1: Clean up build artifacts
```
jonasfj, line 91: Why don't we just make a single bash script for all these steps, that's much easier to debug.

jonasfj (commented):
I haven't tested this, but it looks reasonable to me.

.github/workflows/update-boringssl.yml
HamdaanAliQuatil: Rewrote the script in bash inside the new unified file `tool\bump-boringssl-revision.sh`

.github/workflows/update-boringssl.yml
HamdaanAliQuatil: added a new revision file at `lib/src/third_party/boringssl/REVISION`.

moved this section into the unified script

.github/workflows/update-boringssl.yml
HamdaanAliQuatil: Clubbed everything into a single file `tool\bump-boringssl-revision.sh`
Lets take cleanup of old files to a separate PR after testing this out

jonasfj (approved):
Let's remove the python script, we shouldn't need that right?

jonasfj (dismissed):

lib/src/impl_ffi/impl_ffi.utils.dart
```diff
@@ -319,12 +319,16 @@ extension on _Scope {
...
+    cbs.ref.data = dataAsPointer(data);
+    cbs.ref.len = data.length;
     return cbs;
   }
 
   ffi.Pointer<CBB> createCBB([int sizeHint = 4096]) {
-    final cbb = this<CBB>();
+    // CBB is opaque, so we need to allocate a fixed-size buffer
+    // We can use CBB_init with a reasonable buffer size for the CBB structure
+    final cbb = allocate<ffi.Uint8>(256).cast<CBB>();
```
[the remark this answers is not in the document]
HamdaanAliQuatil: BoringSSL intentionally does not provide an allocation function for `CBB` - it's designed to be stack-allocated in C. There's only `CBB_init()` which initializes an already-allocated structure.

Since we can't stack-allocate in Dart FFI, we could ask the C compiler what the actual size is using sizeof(CBB), and expose that to Dart. The C compiler would know how big `CBB` is on the current platform, and `sizeof(CBB)` would return the actual size in bytes.

lib/src/boringssl/lookup/utils.dart
```diff
@@ -69,9 +69,9 @@ lookupLibraryInDotDartTool() {
 
 /// Find the `.dart_tool/` folder, returns `null` if unable to find it.
 Uri? _findDotDartTool() {
-  // HACK: We have no good mechanism for finding the library created by:
-  //         flutter pub run webcrypto:setup
-  //       So we search relative to the script path and CWD.
+  // HACK: We have no good mechanism for finding the legacy
```
jonasfj, line 72: Is this stuff ever even used anymore? If not let's remove it, but in a different PR.

tool/bump-boringssl-revision.sh
```diff
@@ -0,0 +1,398 @@
+#!/bin/bash
+
+# Copyright 2020 Google LLC
```
jonasfj:
```suggestion
# Copyright 2026 Google LLC
```

tool/bump-boringssl-revision.sh
```diff
@@ -0,0 +1,398 @@
...
+        exit 1
+    fi
+}
+
+mkdirp() {
+    local path="$1"
+    if [[ ! -d "$path" ]]; then
+        mkdir -p "$path"
+    fi
+}
```
jonasfj: what is the harm in just doing `mkdir -p "$path"` instead of having this function wrapping something trivial, I don't think the if condition is necessary

tool/bump-boringssl-revision.sh
```diff
@@ -0,0 +1,398 @@
...
+git_clone_boringssl() {
+    local revision="$1"
+    local temp_dir="$2"
+    local target="$temp_dir/boringssl"
+
+    log_info "Cloning BoringSSL repository..."
+    git clone "$BORINGSSL_REPOSITORY" "$target" >/dev/null 2>&1
+    log_info "Checking out revision: $revision"
+    git -C "$target" checkout --detach "$revision" >/dev/null 2>&1
+    echo "$target"
```
jonasfj: Just take target directory as an argument, instead of taking temp folder as an argument, creating a named directory there and returning it.

tool/bump-boringssl-revision.sh
```diff
@@ -0,0 +1,398 @@
...
+}
+
+cleanup_boringssl() {
+    local path="$ROOT/third_party/boringssl"
+    log_info "Cleaning up old BoringSSL files..."
+    rm -rf "$path"
+    mkdirp "$path"
+}
+
+git_clone_boringssl() {
```
jonasfj, line 137:
```suggestion
# git_clone_boringssl <revision> <destination-folder>
git_clone_boringssl() {
```

May I suggest that we do this for commands/functions that take arguments.

tool/bump-boringssl-revision.sh
```diff
@@ -0,0 +1,398 @@
...
+
+class Capture:
+    def WriteFiles(self, file_sets, asm_outputs):
+        self.file_sets = file_sets
+        self.asm_outputs = asm_outputs
+
+
+boringssl_root = os.path.abspath(sys.argv[1])
+sys.path.insert(0, os.path.join(boringssl_root, "util"))
+import generate_build_files  # type: ignore
```
jonasfj: Do we need to use this? [INCORPORATING.md](https://boringssl.googlesource.com/boringssl/+/HEAD/INCORPORATING.md) says the scirpt is historical.

Can't use the `sources.json` or even better the `gen/sources.cmake`:
https://boringssl.googlesource.com/boringssl/+/HEAD/gen/sources.cmake


--------------

I mean, I don't mind it if we are copying more of boringssl into the repo than we strictly need. We can maybe do a few simple globs to avoid copying the test files and such.

jonasfj (commented):
I don't strictly oppose this approach, but I think we should explore what it would take to use the `gen/sources.cmake` file that is pregenerated and maintained in the boringssl repository.

tool/bump-boringssl-revision.sh
HamdaanAliQuatil: Addressed this in the latest revision.

jonasfj:
So this doesn't work, it doesn't copy the sources in correctly.

tool/bump-boringssl-revision.sh
```diff
@@ -0,0 +1,467 @@
...
+    update_revision "$TARGET_REVISION"
+
+    section "Getting Dart dependencies"
+    log_info "Running 'dart pub get --no-example'..."
+    cd "$ROOT"
+    dart pub get --no-example
+
+    section "Generating symbols table"
+    log_info "Running generate_symbols_table.dart..."
+    dart run "$DIR/generate_symbols_table.dart"
```
jonasfj:
```suggestion
    dart "$DIR/generate_symbols_table.dart"
```

then it won't trigger our own hooks, I think.

jonasfj (approved):

Issue #207, opened 2025-08-06
Could provide a version adapted for Android 15 with a 16KB size limit?


wyqlxf:
problem: Starting November 1, 2025, all new apps and updates to existing apps submitted to Google Play that target Android 15 and higher must support a 16 KB page size on 64-bit devices.

https://developer.android.com/guide/practices/page-sizes?hl=zh-cn

jonasfj:
If someone has suggestions for how to fix this in the build configuration, we're open to PRs.

It'll be a while before I have time to dive into any of this.

mrcsh:
I am not sure not sure why you would still be having issues, but make sure to do a gradle clean and a flutter clean?
and that the NDK is downloaded and installed...

From all the docs I could find it seems that building the .so with the NDK (>=28) is really the only requirement.

marmelad52:
> I am not sure not sure why you would still be having issues, but make sure to do a gradle clean and a flutter clean? and that the NDK is downloaded and installed...
> 
> From all the docs I could find it seems that building the .so with the NDK (>=28) is really the only requirement.

@mrcsh Thank you! Your solution worked perfectly for me. 

Here's what I did to make it work:

1. Updated AGP from 8.12.2 to 8.13 in `settings.gradle`
2. Set NDK to 28.2.13676358 in app's `build.gradle` 
3. Added the `subprojects` configuration to root `build.gradle` as you suggested
4. **Important**: Had to add `kotlinOptions` configuration to avoid JVM-target compatibility issues:
```gradle
if (project.hasProperty('kotlin')) {
    project.tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
        kotlinOptions {
            jvmTarget = "17"
        }
    }
}
```
5. Ran flutter clean and ./gradlew clean as you recommended

PR #212, opened 2025-08-25
Android supports 16KB page size

wyqlxf:
Add 16 KB page size compatibility
https://developer.android.com/guide/practices/page-sizes?hl=zh-tw
#207

### DIFF
--- a/android/CMakeLists.txt
+++ b/android/CMakeLists.txt
@@ -36,11 +36,6 @@ include(
 # https://boringssl.googlesource.com/boringssl/+/HEAD/BUILDING.md#binary-size
 add_definitions(-DOPENSSL_SMALL)
 
-# Add 16KB page size support using Android NDK r27 official method
-# This ensures the library is compatible with 16KB page size systems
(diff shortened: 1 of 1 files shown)


wyqlxf:
@emersion Thank you for the excellent suggestion! I've updated the implementation to follow the official Android guidelines you referenced.

emersion:
Nice! I wonder, do we need to set it twice (one time in the CMake file, another time in the Gradle file)? Isn't one time enough?

seriousGeorgeFX:
Doesn't have the build.gradle be adapted further? I only passed the warning in google play console after additionally adjusting this in the webcrypto build.gradle:

 dependencies {
       classpath 'com.android.tools.build:gradle:8.6.0'
        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:2.2.0"
    }
    
  in "android": setting ndkVersion '28.2.13676358' and compileSdkVersion 36
  in "defaultconfig": setting minSdkVersion 23
  
After that, building a release bundle with your pullrequest branch included in my project as a dependency override, finally the warning in the play console went away.

PR #232, opened 2025-10-13
X25519 and Ed25519 ffi and js implementations

tzaid:
Didn't feel like waiting for v1.1 to get these so I implemented them. The PR includes interfaces and implementations for ffi, js and stubs for both X25519 and Ed25519. For Ed25519, only signBytes and verifyBytes is supported but happy to provide a stream wrapper if required.


jonasfj:
Thanks, I suspect we might still want to polish what we have an get a new version released, before we expand the scope.

Also what is the browser support status, have all agreed implement these yet?

PR #233, opened 2025-10-22
Move PBKDF2 derive bits off the main-thread with Isolates

HamdaanAliQuatil:
Part of https://github.com/google/webcrypto.dart/issues/198

This PR offloads computationally expensive PBKDF2 key derivation operations to separate isolates, preventing them from blocking the main isolate. This follows the same pattern already established for RSA key generation operations.


emersion (approved):
Yup, makes sense to me as well!

Issue #226, opened 2025-09-06
CI: Switch Android emulator from x86 to x86_64 (API 33+) for Flutter


HamdaanAliQuatil:
Raw CI log shows Flutter flags the emulator as “unsupported”.

Our Android CI job boots an x86 (32-bit) emulator on API 28. [Recent Flutter toolchains](https://docs.flutter.dev/release/release-notes/release-notes-3.29.0) don’t support 32-bit x86 emulators, so flutter devices lists the emulator as “unsupported (emulator)” and tests abort before they run.

Evidence (from logs):

```shell
No supported devices found with name or id matching 'android'.

The following devices were found:
Android SDK built for x86 (mobile) • emulator-5554 • unsupported • Android 9 (API 28) (unsupported) (emulator)
```

Our workflow uses `reactivecircus/android-emulator-runner@v2` without overriding arch, whose default is x86 (32-bit). Flutter’s [supported Android architectures](https://docs.flutter.dev/reference/supported-platforms) are x64 (x86_64), Arm32, Arm64 and not 32-bit x86.

Flutter 3.29 [release notes](https://docs.flutter.dev/release/release-notes/release-notes-3.29.0) explicitly removed the android-x86-jit-release artifact, reinforcing that x86 (32-bit) isn’t a viable target.

**Proposed fix:**
In the workflow, explicitly specify a 64-bit emulator architecture and perhaps update the API level as well. 
The `arch` input parameter provided by `reactivecircus/android-emulator-runner@v2`, defaults to x86. We should explicitly choose x86_64 instead.

PR #234, opened 2025-10-23
Workaround missing Android image in CI and upgrade dependencies

emersion:
Google hasn't published an AOSP version of the Android system image for API level 36 yet:
https://issuetracker.google.com/issues/432143095

Temporarily switch to the Google APIs variant to workaround this issue.


emersion:
Ah, yes. Unfortunately this is tricky to debug since workflows in my PRs aren't run by default and need manual action from project members.

I think missing Vulkan/X11 libs are a red herring, and I think the error is caused by a device name mismatch: `android` doesn't exist, `android-x64` does. I've updated my PR to use the correct device name.

emersion:
Added more commits to modernize dependencies.

HamdaanAliQuatil:
Thank you! merging this. Closes #226

HamdaanAliQuatil (approved):

PR #236, opened 2025-11-16
chore: adopt dart native build hooks

### LINKED ISSUE #192: Use Dart native build hooks for packaging
While native hooks in the Dart and Flutter SDK are still experimental, we should explore using them for building.

This would allow this package to no longer be a Flutter plugin, and thus, enable server applications using the Dart SDK to rely on this package.

HamdaanAliQuatil:
Closes #192


jonasfj:
This is awesome! But we probably need to fix lints and get the tests to pass :D

pubspec.yaml
```diff
@@ -37,7 +37,7 @@ dev_dependencies:
   ffigen: ^11.0.0
   test: ^1.16.0
   yaml: ^3.0.0
-  flutter_lints: ^4.0.0
+  flutter_lints: ^6.0.0
```
emersion: Note, this is already done in https://github.com/google/webcrypto.dart/pull/202

hook/build.dart
```diff
@@ -0,0 +1,253 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:ffi';
+import 'dart:io';
+
+import 'package:code_assets/code_assets.dart';
+import 'package:hooks/hooks.dart';
```
jonasfj: We shouldn't consider using something like: https://pub.dev/packages/native_toolchain_cmake

hook/build.dart
```diff
@@ -0,0 +1,131 @@
...
+    // Skip build for non-code targets (e.g. web builds via flutter drive).
+    if (!input.config.buildCodeAssets) {
+      stdout.writeln(
+        'webcrypto: skipping native asset build (code assets not requested).',
+      );
+      return;
+    }
+
+    final targetOs = input.config.code.targetOS;
+    if (!_hostSupports(targetOs)) {
```
jonasfj: So we don't support building for Android?

hook/build.dart
```diff
@@ -0,0 +1,131 @@
...
+  switch (targetArch) {
+    case Architecture.arm64:
+      return const {
+        Abi.androidArm64,
+        Abi.fuchsiaArm64,
+        Abi.iosArm64,
+        Abi.linuxArm64,
+        Abi.macosArm64,
+        Abi.windowsArm64,
+      }.contains(Abi.current());
```
jonasfj: What is Abi.current()?

hook/build.dart
```diff
@@ -0,0 +1,131 @@
...
+            '${installDir.toFilePath()}',
+      );
+    }
+
+    for (final dependency in [
+      packageRoot.resolve('src/CMakeLists.txt'),
+      packageRoot.resolve('src/webcrypto.c'),
+      packageRoot.resolve('src/webcrypto.h'),
+      packageRoot.resolve('src/symbols.generated.c'),
+      packageRoot.resolve('third_party/boringssl/sources.cmake'),
```
jonasfj: We have way more dependencies? What is this?

hook/build.dart
HamdaanAliQuatil: Abi.current() is the ABI of the Dart process running the hook. 

I had used it as a host-architecture guard, but that was the wrong abstraction here since the actual target already comes from input.config.code. Ive removed it.

hook/build.dart
HamdaanAliQuatil: Those are hook invalidation dependencies, not linker dependencies.

I was using a small hand-maintained list of obvious inputs, but that seems too ad hoc. I've changed it to collect actual native build inputs from src/ and third_party/boringssl/ instead.

hook/build.dart
HamdaanAliQuatil: I had added a guard in the hook that restricted it to desktop targets while testing the VM path, and I failed to remove it afterward. I’ve removed that restriction now.

Android still passes in CI after that change. The Android log also shows native build-related activity, including native_toolchain_cmake being resolved and a rebuild after File modified during build.

.github/workflows/test.yml
```diff
@@ -5,6 +5,10 @@ on:
   pull_request:
     branches: [master]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
jonasfj, line 10: where does this come from and how is it related?

.github/workflows/test.yml
jonasfj, line 1: Why do we need to change CI in this PR?

If we need to change CI, can we do it in an independent PR that discusses why are making those changes.

.github/workflows/test.yml
```diff
@@ -38,60 +42,19 @@ jobs:
           sudo apt-get install -y ninja-build libgtk-3-dev
           flutter config --no-analytics
       - run: flutter pub get
-      - run: flutter pub run webcrypto:setup
-      - run: flutter test
+      - run: flutter test --coverage
       - run: flutter test --platform chrome
-      - run: flutter test --platform chrome --wasm
+      - run: dart test -p chrome -c dart2wasm
```
jonasfj: This very cool, now we can test with `dart test` too.

.github/workflows/test.yml
```diff
@@ -38,60 +42,19 @@ jobs:
...
+      - run: xvfb-run dart test -p firefox
       - run: xvfb-run flutter test integration_test/webcrypto_test.dart -d linux
         working-directory: ./example
-      - uses: nanasess/setup-chromedriver@v2
-      - name: Run integration_test with chromedriver
-        working-directory: ./example
-        run: |
-          xvfb-run ../tool/with-chromedriver.sh flutter drive \
```
jonasfj: I'm guessing this stopped working, but that should land in a different PR.

hook/build.dart
```diff
@@ -0,0 +1,124 @@
...
+  await build(args, (input, output) async {
+    // Skip build for non-code targets (e.g. web builds via flutter drive).
+    if (!input.config.buildCodeAssets) {
+      stdout.writeln(
+        'webcrypto: skipping native asset build (code assets not requested).',
+      );
+      return;
+    }
+
+    final packageRoot = Directory.fromUri(input.packageRoot).uri;
```
jonasfj:
```suggestion
    final packageRoot = input.packageRoot;
```

lib/src/boringssl/lookup/lookup.dart
```diff
@@ -26,8 +26,67 @@ import 'utils.dart';
...
+    return null;
+  }
+
+  return (String symbolName) {
+    final sym = symFromString(symbolName);
+    return _nativeWebcryptoLookupSymbol(sym.index);
+  };
+}
+
+_VoidLookup _createLegacyLookup() {
```
jonasfj, line 89: Do we need to keep this around? same with `bin/setup.dart`?
I'm asking is there a need for these? Are native assets not available during tests or something like that?


Also we should update `CHANGELOG.md`

hook/build.dart
```diff
@@ -0,0 +1,124 @@
...
+  '.cc',
+  '.cmake',
+  '.cpp',
+  '.h',
+};
+
+Iterable<Uri> _buildDependencies(Uri packageRoot) sync* {
+  yield* _filesForBuild(
+    Directory.fromUri(packageRoot.resolve('src/')),
+    excludeDirectories: {'build'},
```
jonasfj: Why? are we writing build files into sources that sounds scary? will this happen when installed in PUB_CACHE?

hook/build.dart
```diff
@@ -0,0 +1,124 @@
...
+  for (final entity in root.listSync(recursive: true, followLinks: false)) {
+    if (entity is! File) {
+      continue;
+    }
+    final relativeSegments = entity.uri.pathSegments.skip(
+      root.uri.pathSegments.length,
+    );
+    if (relativeSegments.any(excludeDirectories.contains)) {
+      continue;
+    }
```
jonasfj: make utility methods with tests for this, or find a smarter way.

hook/build.dart
```diff
@@ -0,0 +1,124 @@
...
+    if (!_buildDependencyExtensions.contains(_extension(entity.uri.path))) {
```
jonasfj: you can also just ask `_buildDependencyExtensions.any((e) => entity.uri.path.endsWith(e))`

jonasfj (commented):
I've updated coveralls to not complain about small drops in coverage.

lib/src/boringssl/lookup/lookup.dart
HamdaanAliQuatil: I verified the native asset path works in both `dart test` and `flutter test` now.

I kept the legacy lookup path and `bin/setup.dart` in this PR as a compatibility layer because the current `master` still has CI/docs using `flutter pub run webcrypto:setup`. Let's create a follow-up PR that updates the workflows and README together, rather than mixing the deprecation/removal into this branch.

hook/build.dart
HamdaanAliQuatil: It would have with `buildLocal: true`, and that was a bad default here.

I changed the hook to stop using `buildLocal`, so CMake output no longer goes under `src/build/...`; it now stays under the hook output directory instead. That means we’re not writing build artifacts into package sources or into `PUB_CACHE`.

As a side effect, this also fixed the local `flutter test` “File modified during build” behavior I was seeing.

jonasfj (approved):

PR #238, opened 2025-12-03
add linker flags for 16kb page size

matehat:
This fixes a new issue where the Play Store will reject apps where page size are not 16kb, which, after having fixed similar things in our app, required webcrypto to be compiled accordingly.

### Example issue reported in the Play Store

CleanShot 2025-12-02 at 20 58 32@2x


jonasfj (approved):

jonasfj:
@emersion closing this, as we've landed https://github.com/google/webcrypto.dart/pull/238

Let me know if this is still needed.

PR #241, opened 2026-01-18
Use canonical link for all RFC references 

### LINKED ISSUE #189: Use canonical link for all RFC references
We should be consistent in what resources we link to when linking to RFCs.

YATIN072007:
Closes #189

Updating all RFC references in the modified files to use a single, canonical source. All RFC links now point to their canonical URLs, which improves consistency and avoids redirects or unofficial mirrors.


YATIN072007:
@HamdaanAliQuatil 
Sir kindly review...

PR #247, opened 2026-01-27
Explicitly set ndkVersion to Flutter's

### LINKED ISSUE #239: Current version 0.6.0 is not compatible with 16 KB memory page
Though the issue is raised in this issue https://github.com/google/webcrypto.dart/issues/207 but the solution provided there is ineffective.

The issue is still persisting with this configuration even after incorporating solution cited in the above thread.

Kotlin = "2.2.21"
ndkVersion = "29.0.14206865"

AGP = 8.13.1
Gradle = 8.13

sourceCompatibility JavaVersion.VERSION_17
targetCompatibility JavaVersion.VERSION_17

compileSdk = 36

webcrypto = ^0.6.0

Android Studio Otter | 2025.2.1 Patch 1

APK analyzer showing this:
libwebcrypto.so: 4 KB LOAD section alignment, but 16 KB is require

Please update the package.

emersion:
Without this, webcrypto.dart uses AGP's default NDK version, which can be different from Flutter's preferred NDK version. Apart from causing builds to download multiple NDKs needlessly, this also causes APKs to not use 16 KB page sizes.

Tested by building an APK and opening it in Android Studio's APK analyzer.

Closes: https://github.com/google/webcrypto.dart/issues/239

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -28,6 +28,7 @@ android {
     namespace 'dev.google.webcrypto'
 
     compileSdk = flutter.compileSdkVersion
+    ndkVersion = flutter.ndkVersion
 
     // We depend on cmake for the native parts.
     externalNativeBuild {


HamdaanAliQuatil (approved):
thanks for this 🚀

jonasfj (approved):

emersion:
@jonasfj If you have time, it would be great to have a new release with this fix!

emersion:
@jonasfj Gentle ping, would be nice to have this available on the published package - this is blocking Google Play submissions for all Flutter apps using this library :P

olof-dev:
Can confirm that this fixed the issue for me: the `check_elf_alignment.sh` script (see https://developer.android.com/guide/practices/page-sizes) reports 
`lib/arm64-v8a/libwebcrypto.so: ALIGNED (2**14)`
after this change, instead of the `UNALIGNED (2**12)` without the fix. As @emersion writes, Google Play Console flags a policy violation if it is unaligned, so an updated release on pub.dev would be very welcome.

PR #249, opened 2026-02-07
test: cover ECDH deriveBits length boundaries

### LINKED ISSUE #130: Document the maximum length in EcdhPrivateKey.deriveBits for each curve
Documentation for `EcdhPrivateKey.deriveBits` could have bullet point list mentining the maximum length of bits that can be derived from each curve.

**Example**
```md
[length] specifies the length of the derived secret in bits. The maximum length depends on the [EllipticCurve] used in the keys:
...
 * [EllipticCurve.p521] can provide up to 528 bits.
```

See: 

https://github.com/google/webcrypto.dart/blob/b295d8d0a148e056c285a45c3b38b1ae95ccd34a/lib/src/webcrypto/webcrypto.ecdh.dart#L185-L205

-----------------------------

 * [ ] Find maximum length for each curve.
 * [ ] Add tests for each curve + maximum length in deriveBits.
 * [ ] Document maximum length that can be derived for each curve.

harrshita123:
Follow-up to #289.

This change adds explicit ECDH `deriveBits` boundary coverage for each supported curve:

- exercises the exact maximum length in generated cases;
- verifies the documented maximum succeeds; and
- verifies the first invalid bit length is rejected.

The tested limits are 256 bits for P-256, 384 bits for P-384, and 528 bits for P-521.

### DIFF
--- a/lib/src/testing/ecdh/derive_bits.dart
+++ b/lib/src/testing/ecdh/derive_bits.dart
@@ -1,0 +1,70 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:webcrypto/webcrypto.dart';
+import '../utils/detected_runtime.dart';
+import '../utils/utils.dart';
+
+final _cases = [
+  (name: 'P-256', curve: EllipticCurve.p256, maxBits: 256),
+  (name: 'P-384', curve: EllipticCurve.p384, maxBits: 384),
+  (name: 'P-521', curve: EllipticCurve.p521, maxBits: 528),
+];
+
+void main() => tests().runTests();
+
+List<({String name, Future<void> Function() test})> tests() {
+  final tests = <({String name, Future<void> Function() test})>[];
+  void test(String name, Future<void> Function() fn) =>
+      tests.add((name: name, test: fn));
+
+  for (final c in _cases) {
+    if (detectedRuntime == 'safari' && c.curve == EllipticCurve.p521) {
+      continue;
+    }
+
+    test('ECDH: ${c.name} allows maximum deriveBits length', () async {
+      final aliceKeyPair = await EcdhPrivateKey.generateKey(c.curve);
+      final bobKeyPair = await EcdhPrivateKey.generateKey(c.curve);
+
+      final secret = await aliceKeyPair.privateKey.deriveBits(
+        c.maxBits,
+        bobKeyPair.publicKey,
+      );
+
+      check(secret.length == c.maxBits ~/ 8, 'secret length mismatch');
+    });
+
+    test('ECDH: ${c.name} rejects deriveBits larger than maximum', () async {
+      final aliceKeyPair = await EcdhPrivateKey.generateKey(c.curve);
+      final bobKeyPair = await EcdhPrivateKey.generateKey(c.curve);
+
+      var threw = false;
+      try {
+        await aliceKeyPair.privateKey.deriveBits(
+          c.maxBits + 8,
+          bobKeyPair.publicKey,
+        );
+      } on OperationError {
+        threw = true;
+      }
+      check(
+        threw,
+        'Should throw OperationError for deriveBits larger than maximum',
+      );
+    });
+  }
+
+  return tests;
+}
--- a/lib/src/testing/ecdh/derive_bits.dart
+++ b/lib/src/testing/ecdh/derive_bits.dart
@@ -53,7 +53,7 @@ List<({String name, Future<void> Function() test})> tests() {
       var threw = false;
       try {
         await aliceKeyPair.privateKey.deriveBits(
-          c.maxBits + 8,
+          c.maxBits + 1,
           bobKeyPair.publicKey,
         );
       } on OperationError {
--- a/lib/src/testing/ecdh/derive_bits.dart
+++ b/lib/src/testing/ecdh/derive_bits.dart
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 import 'package:webcrypto/webcrypto.dart';
-import '../utils/detected_runtime.dart';
 import '../utils/utils.dart';
 
 final _cases = [
@@ -30,10 +29,6 @@ List<({String name, Future<void> Function() test})> tests() {
       tests.add((name: name, test: fn));
 
   for (final c in _cases) {
-    if (detectedRuntime == 'safari' && c.curve == EllipticCurve.p521) {
-      continue;
-    }
-
     test('ECDH: ${c.name} allows maximum deriveBits length', () async {
       final aliceKeyPair = await EcdhPrivateKey.generateKey(c.curve);
       final bobKeyPair = await EcdhPrivateKey.generateKey(c.curve);
--- a/lib/src/testing/webcrypto/ecdh.dart
+++ b/lib/src/testing/webcrypto/ecdh.dart
@@ -65,7 +65,8 @@ void main() async {
     generateKeyParams: {'curve': curveToJson(EllipticCurve.p256)},
     importKeyParams: {'curve': curveToJson(EllipticCurve.p256)},
     deriveParams: {},
-    maxDeriveLength: 32,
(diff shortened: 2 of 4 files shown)


lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -206,6 +206,10 @@ final class EcdhPrivateKey {
   /// two parties.
   ///
   /// [length] specifies the length of the derived secret in bits.
+  /// The maximum length that can be derived depends on the elliptic curve:
```
HamdaanAliQuatil:
```suggestion
  /// The maximum allowed [length] is determined by the elliptic curve:
```

HamdaanAliQuatil (changes requested):

test/crypto_subtle_test.dart
```diff
@@ -310,6 +310,73 @@ void main() {
...
+      expect(secret.length, equals(32));
+    });
+
+    test('P-256 rejects deriveBits larger than maximum', () async {
+      final aliceKeyPair = await EcdhPrivateKey.generateKey(EllipticCurve.p256);
+      final bobKeyPair = await EcdhPrivateKey.generateKey(EllipticCurve.p256);
+
+      expect(
+        aliceKeyPair.privateKey.deriveBits(257, bobKeyPair.publicKey),
+        throwsA(anyOf(isA<subtle.JSDomException>(), isA<Error>())),
```
jonasfj: Hmm, this implies a different kind of issue.

I think we are supposed to catch JSDomException and make it into an Exception or an Error.

jonasfj: To be clear that is probably an orthogonal issue to this PR and should be fixed separately.

jonasfj:
Please rebase, I think CI mostly works on master branch now.

test/crypto_subtle_test.dart
harrshita123: since it’s kind of a separate thing, I will leave it out of this PR for now so this doesn’t get messy. I can open a new issue for the JSDomException part if needed.

harrshita123:
> Please rebase, I think CI mostly works on master branch now.

I have rebased the branch onto the latest master so the CI tests can run properly.

test/crypto_subtle_test.dart
```diff
@@ -314,6 +314,73 @@ void main() {
       );
     });
   });
+  group('ECDH deriveBits', () {
+    test('P-256 allows maximum deriveBits length', () async {
```
jonasfj: These tests should decidedly not live in `test/crypto_subtle_test.dart` --- this tests the `subtle.window.crypto.subtle` wrapper we have.

I suggest we put them in `lib/src/testing/ecdh/derive_bits.dart` or `test/ecdh_derive_bits_test.dart` (they won't be included in integration tests, but maybe that's okay -- we can always move them later, and getting coverage is probably better).

--------------

I don't mind these tests, but they could also be made much simpler.


```dart

final _cases = [
  (
    name: 'P-256',
    curve: EllipticCurve.p256,
    maxBits: 256,
  ),
  ...
];

void main() {
  for (final c in _cases) {
    test('${c.name} allows maximum deriveBits length', () async {
      final aliceKeyPair = await EcdhPrivateKey.generateKey(c.curve);
      final bobKeyPair = await EcdhPrivateKey.generateKey(c.curve);

      final secret = await aliceKeyPair.privateKey.deriveBits(
        c.maxBits,
        bobKeyPair.publicKey,
      );

      expect(secret.length, equals(c.maxBits / 8));
    });

    test('${c.name} rejects deriveBits larger than maximum', () async {
      final aliceKeyPair = await EcdhPrivateKey.generateKey(c.curve);
      final bobKeyPair = await EcdhPrivateKey.generateKey(c.curve);

      expect(
        aliceKeyPair.privateKey.deriveBits(c.maxBits + 1, bobKeyPair.publicKey),
        throwsA(anyOf(isA<subtle.JSDomException>(), isA<Error>())), 
      );
    });
  }
}
```

jonasfj (changes requested):

harrshita123:
I updated the branch based on the feedback:

- Rebased the PR onto the latest master, which resolved the merge conflict.
- Kept the ECDH deriveBits documentation change focused on the maximum length for each curve.
- Moved the new ECDH deriveBits boundary tests out of crypto_subtle_test.dart and into the package test runner under lib/src/testing/ecdh/derive_bits.dart.
- Simplified the tests to use shared curve cases for P-256, P-384, and P-521.
- Left the JSDomException behavior out of this PR since it was called out as a separate issue.

I also cleaned up the branch history into a single commit and verified it locally with dart format, dart analyze, and dart test test/webcrypto_test.dart -p vm.

lib/src/testing/ecdh/derive_bits.dart
```diff
@@ -0,0 +1,70 @@
...
+    });
+
+    test('ECDH: ${c.name} rejects deriveBits larger than maximum', () async {
+      final aliceKeyPair = await EcdhPrivateKey.generateKey(c.curve);
+      final bobKeyPair = await EcdhPrivateKey.generateKey(c.curve);
+
+      var threw = false;
+      try {
+        await aliceKeyPair.privateKey.deriveBits(
+          c.maxBits + 8,
```
HamdaanAliQuatil: could this use c.maxBits + 1? deriveBits accepts lengths in bits, including non-byte-aligned lengths, so this is the first invalid value. using +8 could miss an off-by-one implementation that accepts the next seven lengths.

HamdaanAliQuatil (changes requested):
these tests are still useful even though https://github.com/google/webcrypto.dart/issues/130 was closed by https://github.com/google/webcrypto.dart/pull/289. the generated cases added there don’t actually hit the documented limits: maxDeriveLength is a bit count, but those calls pass 32/48/66, and the random range excludes its upper bound.

could we correct those calls here by setting both minDeriveLength and maxDeriveLength to 256/384/528 respectively?

since https://github.com/google/webcrypto.dart/issues/130 is already closed, I’d also reframe this as a follow-up to https://github.com/google/webcrypto.dart/pull/289, replace Fixes https://github.com/google/webcrypto.dart/issues/130, and perhaps rename the PR to test: cover ECDH deriveBits length boundaries.

lib/src/testing/ecdh/derive_bits.dart
harrshita123: I have changed this to c.maxBits + 1 so the test now checks the first invalid bit length and catches any off-by-one behavior.

HamdaanAliQuatil (approved):

lib/src/testing/webcrypto/ecdh.dart
```diff
@@ -65,23 +65,26 @@ void main() async {
...
 
   // P-521: up to 528 bits (66 bytes)
   await runner.generate(
     generateKeyParams: {'curve': curveToJson(EllipticCurve.p521)},
     importKeyParams: {'curve': curveToJson(EllipticCurve.p521)},
     deriveParams: {},
-    maxDeriveLength: 66,
+    minDeriveLength: 528,
+    maxDeriveLength: 528,
   );
```
jonasfj, line 88: I think these `runner.generate` things spit out JSON that we put into `_testData`.

Feel free to make a follow up PR, where we generate a bunch of these with the configurations you put in here. I've usually generated in chrome/firefox/boringssl, but just have one or two of those configurations could be useful.

Mostly those `_testData` entries serves as sanity tests and pins behavior.

lib/src/testing/ecdh/derive_bits.dart
```diff
@@ -0,0 +1,70 @@
...
+
+void main() => tests().runTests();
+
+List<({String name, Future<void> Function() test})> tests() {
+  final tests = <({String name, Future<void> Function() test})>[];
+  void test(String name, Future<void> Function() fn) =>
+      tests.add((name: name, test: fn));
+
+  for (final c in _cases) {
+    if (detectedRuntime == 'safari' && c.curve == EllipticCurve.p521) {
```
jonasfj: Don't ship without a comment on why?

I think you've already fixed it with:
https://github.com/google/webcrypto.dart/pull/352

So probably this actually does work? and we should remove this skip?

jonasfj (approved):

lib/src/testing/ecdh/derive_bits.dart
harrshita123: Since #352 fixed the underlying Safari issue, I’ve removed the P-521 skip and the unused runtime import. The P-521 boundary cases will now run on Safari as well.

lib/src/testing/webcrypto/ecdh.dart
harrshita123: Thanks for clarifying how runner.generate is intended to be used. I’ll keep that work separate from this PR and add a focused follow-up with generated maximum-length vectors. P-521 already has 528-bit vectors, so the follow-up will add the missing 256-bit P-256 and 384-bit P-384 BoringSSL cases.

Issue #60, opened 2022-11-02
Ensure CBS is empty after importing keys


jonasfj:
Whenever we create a CBS object, it's usually because we're feeding some data into BoringCrypto. If all the data is not read, this usually indicates invalid data. If importing a key doesn't consume all the bytes supplied in the key, we should probably reject it.

This needs test cases to ensure compatibility with web implementations.

Look code referencing`Scope.createCBS` those places probably have to check that CBS is empty after it's been processed.

This requires testing of invalid key bytes, so probably blocked by https://github.com/google/webcrypto.dart/issues/55

PR #251, opened 2026-02-09
Reject imported EC and RSA keys with trailing bytes

harrshita123:
This change ensures that all bytes supplied during key import are consumed by BoringSSL.

EC and RSA private/public key imports now verify that the CBS is empty after parsing, rejecting inputs with trailing bytes. This resolves the TODO in EC public key import and aligns behavior with WebCrypto implementations.

Tests for malformed key bytes are not included here, as generating invalid key material is currently blocked by #55.

Related to #60.


lib/src/impl_ffi/impl_ffi.ec_common.dart
```diff
@@ -89,8 +89,14 @@ void _validateEllipticCurveKey(_EvpPKey key, EllipticCurve curve) {
 
 _EvpPKey _importPkcs8EcPrivateKey(List<int> keyData, EllipticCurve curve) {
   return _Scope.sync((scope) {
-    final k = ssl.EVP_parse_private_key(scope.createCBS(keyData));
-    _checkData(k.address != 0, fallback: 'unable to parse key');
+    final cbs = scope.createCBS(keyData);
+    final k = ssl.EVP_parse_private_key(cbs);
+
+    _checkData(
+      k.address != 0 && cbs.ref.len == 0,
```
jonasfj: We need tests that confirm that this what all the browser implementations also do.

I wonder if we should simply write some hand written tests? that don't necessarily use the testrunner setup..

Maybe, we could do something like:

`lib/src/testing/regression/issue_<id>_<name>.dart`

And have such files do:
```
import '../utils/test.dart';

void runTests(void Function(String name, Future<void> Function() test) testFn) {
  test('something', () async {
    ...
  });
}
```

Then we can have `lib/src/testing/testing.dart` import and call `runTests` in its `runAllTests` function.

@HamdaanAliQuatil WDYT?

HamdaanAliQuatil: I like the idea of an issue-specific regression bucket.

Since `runAllTests` already aggregates non-`TestRunner` tests like `random.dart` and `digest.dart`, I’d follow that existing pattern and add something like `lib/src/testing/regression/issue_60_trailing_bytes.dart` exposing `tests()`, then append those in `lib/src/testing/testing.dart`.

That said, I wouldn’t want `regression/issue_<id>_...` to become the default pattern for lots of tests. I’d treat it as a narrow escape hatch for browser-compat regressions and other negative cases that don’t cleanly fit the current framework.

lib/src/testing/regression/issue_60_trailing_bytes.dart
```diff
@@ -0,0 +1,84 @@
...
+    try {
+      await EcdsaPublicKey.importSpkiKey(invalidKey, EllipticCurve.p256);
+    } on FormatException {
+      threw = true;
+    }
+    check(threw, 'Expected FormatException when importing EC public key with trailing bytes');
+  });
+
+  test('RsaPss: importPkcs8Key with trailing bytes throws FormatException', () async {
+    final kp = await RsaPssPrivateKey.generateKey(2048, BigInt.from(65537), Hash.sha256);
```
jonasfj: We could also hardcode these keys, to make tests faster and more consistent.

Maybe it might also be worth testing that before we append zero, we can't import, and when we've appended zero we can't import. This way we're sure it's not because we have hardcoded an invalid :D

lib/src/testing/regression/issue_60_trailing_bytes.dart
```diff
@@ -0,0 +1,84 @@
...
+import 'package:webcrypto/webcrypto.dart';
+import '../utils/utils.dart';
+
+void main() => tests().runTests();
+
+/// Tests for issue #60, exported for use in `../testing.dart`.
+List<({String name, Future<void> Function() test})> tests() {
+  final tests = <({String name, Future<void> Function() test})>[];
+  void test(String name, Future<void> Function() testFn) =>
+      tests.add((name: name, test: testFn));
```
jonasfj: We could add something like this to `utils/utils.dart`
```dart
List<({String name, Future<void> Function() test})> defineTests(
  void Function(void Function(String name, Future<void> Function() testFn) test) tests
) {
  final tests_ = <({String name, Future<void> Function() test})>[];
  void test(String name, Future<void> Function() testFn) =>
      tests_.add((name: name, test: testFn));
  tests(test);
  return tests_;
}
```

Then this could just be:

```dart
void main() => tests.runTests();

final tests = defineTests((test) {
  test('...', () async {
    ...
  });

  ...
});
```

But we don't have to do that in this PR. I'm happy to land this as is.

jonasfj (approved):

lib/src/impl_ffi/impl_ffi.ec_common.dart
jonasfj: I'd love to fit as much as possible into `TestRunner` -- and I do think `TestRunner` (or something like it) could be used to test exceptions, but fact is also that `TestRunner` is a bit hard to keep track of today.

It's complicated, and adding support for testing a new kind of thing is just hard, very hard.


So maybe it's fine that we add more one-off test files like this. Also because otherwise how are we going to land these things we can't just land without tests, and we can't wait for `TestRunner` to support more features.

Also no matter how far we go with `TestRunner` it'll never be able to cover everything.

jonasfj (approved):

lib/src/testing/regression/issue_60_trailing_bytes.dart
harrshita123: > Done. I replaced generated keys with hardcoded EC/RSA PKCS8 and SPKI keys to make the tests faster
  > and deterministic. Each test now first imports the valid key successfully, then appends a trailing
  > byte and verifies that import is rejected.

lib/src/testing/regression/issue_60_trailing_bytes.dart
harrshita123: > Yes, the requested regression test coverage is added and wired into runAllTests. I also updated the
  > tests to use hardcoded keys and verify the original valid key imports before checking rejection
  > with trailing bytes. I left the defineTests helper out for now since that was marked as optional
  > for this PR.

lib/src/testing/regression/issue_60_trailing_bytes.dart
```diff
@@ -0,0 +1,102 @@
...
+import 'dart:typed_data';
+import 'package:webcrypto/webcrypto.dart';
+import '../utils/utils.dart';
+
+void main() => tests().runTests();
+
+const _isWeb = bool.fromEnvironment('dart.library.js') ||
+    bool.fromEnvironment('dart.library.wasm');
+
+const _ecPkcs8 = 'MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQgYLo96dEstESlXmgZWdeQoLYOSpUzxzaTFzdc/KtJzoGhRANCAATdpa7vSPqEDzyhDB3JIXi8NL5pByNzLMxq9CirGP9PzaxiW58t+I0KzUNUwy0McBdzuYAKBFPhGgjzvYJxIExE';
```
jonasfj: nit:

```suggestion
const _ecPkcs8 = '''
    MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQgYLo96dEstESlXmgZW
    deQoLYOSpUzxzaTFzdc/KtJzoGhRANCAATdpa7vSPqEDzyhDB3JIXi8NL5pByNzLM
    xq9CirGP9PzaxiW58t+I0KzUNUwy0McBdzuYAKBFPhGgjzvYJxIExE
  '''.replace(RegExp(r'\s+'), '');
```

or something like that if you want to break the lines.

lib/src/testing/regression/issue_60_trailing_bytes.dart
```diff
@@ -0,0 +1,102 @@
...
+
+import 'dart:convert';
+import 'dart:typed_data';
+import 'package:webcrypto/webcrypto.dart';
+import '../utils/utils.dart';
+
+void main() => tests().runTests();
+
+const _isWeb = bool.fromEnvironment('dart.library.js') ||
+    bool.fromEnvironment('dart.library.wasm');
```
jonasfj:
```suggestion
const _isWeb = bool.fromEnvironment('dart.library.js_util');
```

I think that does it.

lib/src/testing/regression/issue_60_trailing_bytes.dart
```diff
@@ -0,0 +1,102 @@
...
+
+    // Verify key with trailing bytes is rejected.
+    bool threw = false;
+    try {
+      await EcdsaPrivateKey.importPkcs8Key(badKey, EllipticCurve.p256);
+    } on FormatException {
+      threw = true;
+    }
+    if (!_isWeb) {
+      check(threw, 'Should throw FormatException for trailing bytes');
```
jonasfj: What happens that is different on web?

harrshita123: On web, `crypto.subtle.importKey()` rejects the malformed PKCS8/SPKI input with trailing bytes, and our JS wrapper maps that browser `DataError` to `FormatException`. The FFI backend was previously more permissive because BoringSSL parsed the valid prefix and left the trailing bytes unread. This change makes FFI match the browser behavior by requiring the input to be fully consumed.

harrshita123:
@jonasfj 
The regression test now gates the PKCS8 trailing-byte assertion on the FFI backend using dart.library.ffi, since this PR changes the BoringSSL/FFI parser behavior. SPKI trailing-byte checks still run across platforms. This should avoid the Firefox Dart2Wasm failure while keeping coverage for the implementation change.

lib/src/impl_ffi/impl_ffi.ec_common.dart
HamdaanAliQuatil: @harrshita123 here malformed imports leak native EVP_PKEY objects.

parsing may return a valid allocated key while leaving trailing bytes. `_checkData` then throws before `_EvpPKey.wrap(k)` registers its finalizer, leaking the key. Repeated malformed imports could exhaust memory.

it should look something like:
```dart
_checkData(k.address != 0, fallback: 'unable to parse key');
final key = _EvpPKey.wrap(k);
_checkData(cbs.ref.len == 0, fallback: 'unable to parse key');
```

apply this to all four EC/RSA PKCS8/SPKI paths.

lib/src/testing/regression/issue_60_trailing_bytes.dart
HamdaanAliQuatil: @harrshita123 PKCS8 browser regression tests silently assert nothing here. 

assertions are guarded by `_isFfi`. on web platforms, the import runs but either outcome passes. This hides the Firefox Dart2Wasm incompatibility that prompted the guard and does not establish the claimed browser parity. 

either make these explicitly FFI-only tests or document and test backend-specific behavior.

HamdaanAliQuatil (changes requested):

lib/src/impl_ffi/impl_ffi.ec_common.dart
harrshita123: I hadn't considered the case where parsing succeeds but the trailing-bytes validation throws before the EVP_PKEY is wrapped and its finalizer is registered.

I've updated the EC/RSA import paths so the parsed key is wrapped immediately after validating the returned pointer, and the trailing-bytes check is performed afterward. This should prevent native key objects from being leaked when malformed inputs are rejected.

lib/src/testing/regression/issue_60_trailing_bytes.dart
harrshita123: the previous test effectively asserted nothing on web backends, so it didn't establish the browser parity that the comment claimed.

I've revised the regression tests to make the backend-specific behavior explicit and removed the silent pass-through on web. This should make the intent of the test much clearer.

jonasfj (approved):

HamdaanAliQuatil (approved):

PR #253, opened 2026-02-14
ci: reduce test matrix and eliminate redundant runs

HamdaanAliQuatil:
part of #255 

- remove macos-14 job (identical to macos-15, doubles CI time)
- merge linux-coverage and windows-coverage into main jobs (re-ran all tests)
- fix redundant test runs: VM and Chrome were tested 2-3x per job
- add concurrency group to cancel superseded runs on same PR branch
- make iOS/Android continue-on-error so emulator flakiness doesn't block

reduces from 10 jobs to 7. fewer test executions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,10 @@ on:
   pull_request:
     branches: [master]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 env:
   PUB_ENVIRONMENT: bot.github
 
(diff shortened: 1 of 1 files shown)


jonasfj:
> make iOS/Android continue-on-error so emulator flakiness doesn't block

I don't think this means ignore errors. it just means continue running steps, right?

HamdaanAliQuatil:
> make iOS/Android continue-on-error so emulator flakiness doesn't block
> > I don't think this means ignore errors. it just means continue running steps, right?

I originally meant "keep the rest of the matrix running even if an emulator job flakes", but `continue-on-error` would make an exhausted iOS/Android failure non-blocking, which is too weak for CI.

After rebasing on top of #257, I'm not keeping this part. iOS/Android now use retry/timeouts instead, and they still fail the workflow if all attempts fail.

jonasfj (approved):

Issue #255, opened 2026-02-17
Improve CI reliability and efficiency


HamdaanAliQuatil:
CI is unreliable: tests fail randomly. Multiple issues:

- Tests run 2-3 times per job (VM twice, Chrome twice)
- Separate coverage jobs duplicate all test runs
- macOS times out after 15 minutes
- Mobile emulators are flaky

We plan to ship it in two parts:

Structural Cleanup: 
- [x] Add concurrency group (cancel superseded runs)
- [x] Remove macos-14 (duplicate of macos-15)
- [x]  Merge linux-coverage and windows-coverage into main jobs
- [x]  Fix redundant test runs (VM/Chrome now run once per job)
- [ ]  ~~Make iOS/Android non-blocking (continue-on-error: true)~~

Emulator Reliability:
The goal is to make mobile tests deterministic by enabling retries and improving infrastructure.

Android

- [ ]  ~~Downgrade API 36 to 34 (36 has a known emulator issue; need to evaluate if it was upgraded for a specific reason)~~
- [ ]  Cache AVD snapshots (actions/cache@v4)
- [x]  Add headless flags: -no-window -gpu swiftshader_indirect -noaudio
- [x]  Add 3-attempt retry loop 

iOS
- [ ]  Add 2-attempt retry wrapper (nick-fields/retry@v3)

Future Discussion:

- [ ]  Debug macOS WASM hang (causes 15min timeout)
- [ ]  Evaluate if Firefox is needed on all platforms
- [ ]  Consider Linux-only coverage

Open for opinions

harrshita123:
Hi ,
I am interested in working on this PR and would love to help move it forward. Please let me know if there is anything specific I can take up or assist with.

HamdaanAliQuatil:
Hey @harrshita123, sure, we'd love some help!

For context, we are currently migrating to Dart-native build hooks. I've updated several tests in #236 that need to ship alongside those native hooks, so those specific test updates are currently blocked until the migration lands.

There are several CI infrastructure fixes we can land immediately. When I say the tests "fail randomly," here is exactly what's happening under the hood:

1. Test Timeouts: You might notice `RSA-PSS: 4096/e3/sha-384/s0 -- generateKeyPair()` timing out. RSA-4096 key generation is computationally expensive and frequently exceeds the default 30-second test timeout on loaded CI runners.
[...]

PR #257, opened 2026-03-08
ci: retry flaky iOS tests and run Android emulator headless

harrshita123:
Part of #255

iOS was failing due to the simulator debug connection timing out, so wrapped the test step with nick-fields/retry (2 attempts).

For Android added the headless emulator flags (-no-window -gpu swiftshader_indirect  -no audio -no-boot-anim).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,14 +43,6 @@ jobs:
       - run: flutter test --platform chrome
       - run: xvfb-run flutter test integration_test/webcrypto_test.dart -d linux
         working-directory: ./example
-      - uses: nanasess/setup-chromedriver@v2
-      - name: Run integration_test with chromedriver
-        working-directory: ./example
-        run: |
-          xvfb-run ../tool/with-chromedriver.sh flutter drive \
-            --driver=test_driver/integration_test.dart \
-            --target=integration_test/webcrypto_test.dart \
-            -d chrome
       - run: xvfb-run flutter pub run test -p vm,chrome,firefox
   macos-14:
     name: MacOS 14 desktop / Chrome / Firefox
@@ -137,15 +129,6 @@ jobs:
       #- run: flutter test --platform chrome --wasm
       - run: flutter test integration_test/webcrypto_test.dart -d windows
         working-directory: ./example
-      - uses: nanasess/setup-chromedriver@v2
-      - name: Run integration_test with chromedriver
-        working-directory: ./example
-        shell: bash
-        run: |
-          ../tool/with-chromedriver.sh flutter drive \
-            --driver=test_driver/integration_test.dart \
-            --target=integration_test/webcrypto_test.dart \
-            -d chrome
       - run: flutter pub run test -p vm,chrome,firefox
   ios:
     name: iOS emulator (iPhone)
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,7 +41,6 @@ jobs:
       - run: flutter pub run webcrypto:setup
       - run: flutter test
       - run: flutter test --platform chrome
-      - run: flutter test --platform chrome --wasm
       - run: xvfb-run flutter test integration_test/webcrypto_test.dart -d linux
         working-directory: ./example
       - uses: nanasess/setup-chromedriver@v2
@@ -70,7 +69,6 @@ jobs:
       - run: flutter pub run webcrypto:setup
       - run: flutter test
       - run: flutter test --platform chrome
-      - run: flutter test --platform chrome --wasm
       - run: flutter test integration_test/webcrypto_test.dart -d macos
         working-directory: ./example
       # TODO: Enable chromedriver testing on MacOS when it works reliably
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -167,8 +167,12 @@ jobs:
         with:
           model: 'iPhone 15'
       - run: flutter pub get
-      - run: flutter test integration_test/webcrypto_test.dart -d iphone
-        working-directory: ./example
+      - name: Run integration tests
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 20
```
HamdaanAliQuatil, line 153: The outer ios job still has `timeout-minutes: 15` above. That means the workflow can be killed by the job timeout before the second attempt has a chance to help. Could you also bump the ios job timeout to match the retry budget?

.github/workflows/test.yml
```diff
@@ -195,13 +199,17 @@ jobs:
...
           arch: x86_64
+          emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim
           working-directory: ./example
-          script: flutter test integration_test/webcrypto_test.dart -d emulator
+          script: |
+            for attempt in 1 2 3; do
+              flutter test integration_test/webcrypto_test.dart -d emulator --timeout 60s && break
+              echo "Attempt $attempt failed, retrying..."
+              [ $attempt -eq 3 ] && exit 1
+            done
```
HamdaanAliQuatil: This retry loop is currently broken in CI. On the PR head run, Android fails with "/usr/bin/sh: 1: Syntax error: end of file unexpected (expecting \"done\")". Could you rewrite this so the emulator action can safely pass it to sh -c?

try 
```
script: >-
bash -lc 'for attempt in 1 2 3; do
....
```

.github/workflows/test.yml
```diff
@@ -195,13 +199,17 @@ jobs:
       - name: Run flutter test integration_test/webcrypto_test.dart -d emulator
         uses: reactivecircus/android-emulator-runner@v2
         with:
-          api-level: 36
-          # TODO: switch back to default once this is issue resolved:
-          # https://issuetracker.google.com/issues/432143095
+          api-level: 34
```
HamdaanAliQuatil: I dug through the history on this before deciding. We had intentionally bumped Android CI to API 36 in #231  to align with Flutter’s recommended setup, and then in #234 we kept API 36 and fixed the CI breakage by switching to the google_apis image plus x86_64/emulator handling.

So I don’t think we should downgrade to 34 here unless we have a current reproducible API-36-specific blocker. Could you keep this at 36 and preserve the retry/headless improvements independently?

HamdaanAliQuatil: also please preserve the comment. we'll need that later for reference.

HamdaanAliQuatil (changes requested):

HamdaanAliQuatil:
@harrshita123 the new dart2wasm smoke job is the right direction. It is failing because two existing tests still expect the old dart2wasm error string, JavaScriptError. 

In the current toolchain, Chrome/Dart2Wasm surfaces the browser DOMException name instead: QuotaExceededError for oversized buffers and TypeMismatchError for non-integer typed arrays. Could you update those two expectations to accept both forms?

harrshita123:
> @harrshita123 the new dart2wasm smoke job is the right direction. It is failing because two existing tests still expect the old dart2wasm error string, JavaScriptError.
> 
> In the current toolchain, Chrome/Dart2Wasm surfaces the browser DOMException name instead: QuotaExceededError for oversized buffers and TypeMismatchError for non-integer typed arrays. Could you update those two expectations to accept both forms?

Updated this. I changed the two getRandomValues expectations so they accept both the older JavaScriptError form and the newer DOMException-name form surfaced by the current dart2wasm toolchain.

HamdaanAliQuatil:
The remaining timeouts are now coming from the chromedriver `flutter drive -d chrome integration` step. On Linux, the test prints All tests passed! and then hangs during teardown. On Windows, DWDS fails to establish the Chrome debug connection and then hangs until the job timeout.

Could you remove the chromedriver `flutter drive -d chrome` steps from both the Linux and Windows desktop jobs? We still retain Chrome browser coverage through `flutter test --platform chrome / flutter` `pub run test -p chrome, desktop` integration coverage through `flutter test integration_test/... -d linux/windows/macos`, Firefox coverage, and the new Linux dart2wasm smoke job. So this removes the flaky duplicate WebDriver path rather than removing unique package coverage.

HamdaanAliQuatil (approved):
lgtm. all tests pass. thanks!

we should add a longer per-test timeout to the native macOS integration step, probably matching Android.

harrshita123:
> lgtm. all tests pass. thanks!
> 
> we should add a longer per-test timeout to the native macOS integration step, probably matching Android.



> lgtm. all tests pass. thanks!
> 
> we should add a longer per-test timeout to the native macOS integration step, probably matching Android.

I’ll update that and add a longer timeout for the native macOS integration step as well.

PR #258, opened 2026-03-19
example: remove and ignore pubspec.lock

kevmoo:


lib/src/impl_js/impl_js.utils.dart
```diff
@@ -131,11 +131,14 @@ Future<T> _handleDomException<T>(
 }) async {
   try {
     return await fn();
-  } on subtle.JSDomException catch (e) {
-    throw _translateDomException(
-      e,
-      invalidAccessErrorIsArgumentError: invalidAccessErrorIsArgumentError,
-    );
+  } catch (e) {
+    if (e.isA<subtle.JSDomException>()) {
```
jonasfj, line 135: how is this better? were we doing it wrong before?

kevmoo: You can't do a normal `on X` check for JS exceptions. This doesn't work in Wasm!

srujzs: It's not just incompatibility with Wasm, it's also extension types. `on X` is a runtime check, and at runtime, `JSDomException` is just some `JSObject` from a JS compiler's perspective and any JS value from dart2wasm's perspective. If you wanted to check that it actually is a `JSDomException`, you'll need to use interop to type-check, which is what `isA` does.

A slightly deeper dive into the differing semantics: https://dart.dev/interop/js-interop/js-types#compatibility-type-checks-and-casts. With a recent enough SDK version, you'll see `on <interop_type>` checks being linted as well.

mosuem:
Closing as most changes are now landed through other PRs.

PR #262, opened 2026-04-20
fix(web): detect secure context for subtle crypto

### LINKED ISSUE #159: Better detection of when `window.crypto.subtle` works (detection of secure context)
The webcrypto API is only available in "secure contexts". We should detect when this isn't the case, make sure we throw `UnsupportedError` with a description error message, and ideally a link to some documentation about "secure contexts".

 * [Webcrypto API docs on mdn](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API)
   * This says:
     > Secure context: This feature is available only in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS), in some or all [supporting browsers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API#browser_compatibility).

 * [Secure Contexts on mdn](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts)

We could potentially write up a note about this in the "compatibility notes" section in the `README.md`.

-------------

Also it'd be nice if we had some test cases for this. Those might have to be manual test cases, because `dart test` running in the browser will probably fake a secure context.

HamdaanAliQuatil:
closes #159

### DIFF
--- a/lib/src/crypto_subtle.dart
+++ b/lib/src/crypto_subtle.dart
@@ -64,7 +64,9 @@ extension type JSWindow(JSObject _) implements JSObject {
 extension type JSCrypto(JSObject _) implements JSObject {
   /// https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle
   @JS('subtle')
-  external JSSubtleCrypto? get subtleOrNull;
+  external JSSubtleCrypto? get _subtleOrNull;
+
+  JSSubtleCrypto get subtle => _requireSubtleCrypto();
 
   /// https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
   external JSTypedArray getRandomValues(JSTypedArray array);
@@ -75,8 +77,8 @@ const _webCryptoApiDocumentationUrl =
 const _secureContextsDocumentationUrl =
     'https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts';
 
-JSSubtleCrypto requireSubtleCrypto() {
-  final subtle = window.crypto.subtleOrNull;
+JSSubtleCrypto _requireSubtleCrypto() {
+  final subtle = window.crypto._subtleOrNull;
   if (subtle != null) {
     return subtle;
   }
@@ -423,7 +425,7 @@ Future<ByteBuffer> decrypt(
   JSCryptoKey key,
   Uint8List data,
 ) async {
-  final subtleCrypto = requireSubtleCrypto();
+  final subtleCrypto = window.crypto.subtle;
   final value = await subtleCrypto
       .decrypt(algorithm.toJS, key, data.toJS)
       .toDart;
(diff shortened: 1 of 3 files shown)


lib/src/crypto_subtle.dart
```diff
@@ -60,12 +63,38 @@ extension type JSWindow(JSObject _) implements JSObject {
 /// https://developer.mozilla.org/en-US/docs/Web/API/Crypto
 extension type JSCrypto(JSObject _) implements JSObject {
   /// https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle
-  external JSSubtleCrypto get subtle;
+  @JS('subtle')
+  external JSSubtleCrypto? get subtleOrNull;
```
jonasfj:
```suggestion
  @JS('subtle')
  external JSSubtleCrypto? get _subtleOrNull;
  
  JSSubtleCrypto get subtle => requireSubtleCrypto();
```

Then we don't need to refactor the rest of the code.

We can even make `_requireSubtleCrypto` a private function, or inline it inside the `subtle` getter.

jonasfj (approved):
Minor improvement, but let's consider putting the logic into the `subtle` getter

lib/src/crypto_subtle.dart
HamdaanAliQuatil: Makes sense. This is a good design. I'll move the logic behind `crypto.subtle` and keep the nullable raw getter private, so callers can keep using `window.crypto.subtle` normally.

HamdaanAliQuatil:
Added targeted coverage for the new `crypto.subtle` error paths.

The getter design stays the same; I extracted the branch selection into a testable helper and added tests for the two negative cases. Coveralls increased by 0.02%.

test/crypto_subtle_test.dart
```diff
@@ -150,9 +194,11 @@ void main() {
   });
 
   group('crypto.subtle', () {
+    final cryptoSubtle = subtle.window.crypto.subtle;
+
     test('generateCryptoKey: success', () async {
       expect(
-        await subtle.window.crypto.subtle
+        await cryptoSubtle
```
jonasfj: if we want to make this change, do it in a different PR

HamdaanAliQuatil:
Makes sense. That alias is unrelated cleanup, so I dropped it from this PR and kept this one focused on the secure-context change.

jonasfj (approved):

lib/src/crypto_subtle.dart
```diff
@@ -60,12 +63,44 @@ extension type JSWindow(JSObject _) implements JSObject {
...
   external JSTypedArray getRandomValues(JSTypedArray array);
 }
 
+const _webCryptoApiDocumentationUrl =
+    'https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API';
+const _secureContextsDocumentationUrl =
+    'https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts';
+
+@visibleForTesting
+JSSubtleCrypto requireSubtleCrypto(
```
jonasfj, line 82: For the record, I'd be okay not testing this.

jonasfj: filed: https://github.com/google/webcrypto.dart/issues/263

I'm sure we have other things like this too.

PR #264, opened 2026-04-30
Cleanup all the cruft from being a plugin, now we are a dart package with hooks

jonasfj:
This is an attempt to remove all the plugin left-over and rely exclusively on native hooks.

I had to re-create the example/, because the templates change so much that I couldn't keep up.

everything in example/ is boilerplate flutter except:
 * `lib/main.dart` (copied from the old example)
 * `integration_test/`
 * `pubspec.yaml`


jonasfj:
wow, I did not think I was likely to get CI passing, after all the files I just deleted :see_no_evil:

pubspec.yaml
HamdaanAliQuatil:
```suggestion
description: Cross-platform implementation of Web Cryptography APIs for Dart.
```

:)

tool/test.sh
```diff
@@ -21,18 +21,12 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
...
-flutter pub run webcrypto:setup
-
-section 'flutter test (local)'
-flutter test
-
-
-section 'flutter test (chrome)'
-flutter test --platform chrome
+section 'dart test (vm,chrome,firefox)'
+#xvfb-run dart test -p vm,chrome,firefox
```
HamdaanAliQuatil: Is this intentionally commented out? The script still has a `dart test (vm,chrome,firefox)` section and still ends with `### All tests passed`, but the root package test command is commented out.

I think we should either restore this command or remove the misleading section if `tool/test.sh` is no longer intended to validate the root package.

HamdaanAliQuatil (changes requested):

tool/test.sh
jonasfj: nope that's a bug

jonasfj:
maybe we should disable firefox testing on mac os, it seems to fail a bit intermittently.

And honestly, testing firefox on Linux is probably good enough.

HamdaanAliQuatil (approved):
Looks good!
Nice to finally have `webcrypto` as a Dart package with hooks :)

PR #266, opened 2026-05-01
ci: trim browser matrix and cache Android AVD

HamdaanAliQuatil:
Part of #255

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,11 +49,8 @@ jobs:
       - name: Convert coverage to lcov
         run: dart run coverage:format_coverage -i ./coverage -o ./coverage/lcov.info --lcov --report-on lib/
       - uses: coverallsapp/github-action@v2
-        with:
-          flag-name: linux
-          parallel: true
   macos:
-    name: MacOS desktop / Chrome / Firefox
+    name: MacOS desktop / Chrome
     runs-on: macos-15 # Test with xcode 16
     timeout-minutes: 15
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -49,11 +49,8 @@ jobs:
       - name: Convert coverage to lcov
         run: dart run coverage:format_coverage -i ./coverage -o ./coverage/lcov.info --lcov --report-on lib/
       - uses: coverallsapp/github-action@v2
-        with:
-          flag-name: linux
-          parallel: true
```
jonasfj, line 54: Oh, is this to only report coverage from this job

Imo fine... If that's what it aims to do.

jonasfj (approved):

.github/workflows/test.yml
```diff
@@ -27,74 +27,37 @@ jobs:
       - run: dart format --output none --set-exit-if-changed .
       - run: dart analyze --fatal-warnings .
   linux:
-    name: Linux desktop / Chrome / Firefox
+    name: Linux Dart / Chrome / Firefox
```
jonasfj: I think dart only is new job

.github/workflows/test.yml
```diff
@@ -27,74 +27,37 @@ jobs:
...
-      - name: Configure Flutter
-        run: |
-          sudo apt-get update -y
-          sudo apt-get install -y ninja-build libgtk-3-dev
-          flutter config --no-analytics
-      - run: flutter pub get
+      - uses: dart-lang/setup-dart@v1
+      - run: dart pub get --no-example
       - run: xvfb-run dart test -p vm,chrome,firefox -c dart2js,dart2wasm --coverage ./coverage
-      - run: xvfb-run flutter test integration_test/webcrypto_test.dart -d linux
```
jonasfj, line 46: We want to keep this. It's a build mode separate from the others

.github/workflows/test.yml
HamdaanAliQuatil: taking this to a new PR

jonasfj (approved):

PR #275, opened 2026-05-24
feat: add prebuilt native asset hook support

HamdaanAliQuatil:


prebuilt/README.md
```diff
@@ -0,0 +1,28 @@
+Prebuilt native assets
+======================
+
+This directory is reserved for prebuilt `webcrypto` native libraries that can
+be bundled by `hook/build.dart` without requiring a local C/C++ toolchain.
+
+These binaries should be generated by maintainer-triggered CI from merged
+source, then committed by a maintainer. They should not be accepted from
+user-authored PRs.
```
jonasfj, line 9: I just want to highlight alternative options, in no particular order.

* (A) We build `prebuilt/` as part of the automated publishing job, they are always build github actions and published inside the package to pub.dev
* (B) We build `prebuilt/` and upload as artifacts on the github release we create when we create a new version, the hook will then have to download these from github releases.
* (C) We build `prebuilt/` and then commit them to git.

(C) has the downside that repository size will increase significantly over time, especially if we add more architectures/platforms.

(A) has the downside that download size of users will increase, but not that much. pub.dev will refuse packages larger than 100mb or 200mb (I don't recall the exact limits), but we'll probably be okay unless we add many architectures.

--------

If we can build the artifacts in github CI, then perhaps (A) is attractive? We also avoid maintainer churn and avoid having to store binary artifacts in github.

mosuem:
@HamdaanAliQuatil I filed #325, based on `package:icu4x`s [existing infrastructure](https://github.com/unicode-org/icu4x/blob/main/ffi/dart/hook/build.dart). Would you be fine with leaving a review on that PR and me closing this one?

HamdaanAliQuatil:
Thanks for picking this up @mosuem . Yes, please close this PR since the implementation here is stale.

I resumed the work in #315 after Jonas’s feedback and moved prebuilt generation into the existing pub.dev publishing workflow. #325 uses the alternative GitHub Release fetch model, so it does not directly supersede #315. I’ll leave a review there so we can settle which distribution model we want before either implementation goes further.

PR #279, opened 2026-05-29
refactor: replace runtime symbol lookup with direct FFI  bindings

### LINKED ISSUE #277: FFI refactor: reduce/remove indirect symbol lookup
Today most of the native BoringSSL surface is reached indirectly through:

  - `src/symbols.yaml`
  - `tool/generate_symbols_table.dart`
  - `src/symbols.generated.c`
  - `lib/src/boringssl/lookup/symbols.generated.dart`
  - `lib/src/boringssl/lookup/lookup.dart`
  - `lib/src/third_party/boringssl/generated_bindings.dart`

  The core pattern is:

  - export one `webcrypto_lookup_symbol(index)` function from `src/webcrypto.c`
  - map symbol name -> enum -> index in Dart
  - resolve function pointers at runtime
  - call most of BoringSSL through `BoringSsl.fromLookup(lookup)`

This solved symbol-conflict concerns, but it has some downsides:

  1. It hides actual symbol usage from the linker/toolchain.
  2. It makes native tree shaking / dead stripping much harder.
  3. It keeps a fair amount of custom codegen and symbol-table machinery alive.
  4. It makes the native interface harder to reason about than direct bindings.

The single generic lookup function likely makes tree shaking effectively impossible in the current shape.

## Goal

  Rework the FFI layer so that more of the native surface is explicit and direct.

  Possible directions:

  - move practical parts of the surface to direct `@Native(..., assetId: ...)` bindings
  - reduce dependence on `BoringSsl.fromLookup(lookup)`
  - keep wrapper/helper exports only where they are actually needed
  - if symbol conflicts are still a concern, evaluate prefixing/renaming exported symbols instead of using a generic runtime lookup trampoline

  ## Non-goals

  This issue is not the full prebuilt-assets rollout and not the full static-linking/tree-shaking redesign. It is the architectural cleanup needed before those later steps.

HamdaanAliQuatil:
Closes #277 

This replaces the runtime `webcrypto_lookup_symbol(index)` indirection with ffigen-generated direct `@Native` bindings.

How this works:

  - The native BoringSSL build is compiled with `BORINGSSL_PREFIX=webcrypto`, so the bundled native asset exports
  prefixed symbols such as `webcrypto_EVP_PKEY_new`.
  - ffigen is run with the same prefix configuration and an `ffi-native` asset id of `package:webcrypto/webcrypto.dart`.
  - ffigen generates direct `@Native(..., assetId: ...)` bindings to the prefixed native symbols.
  - A rename rule exposes the Dart API with the original BoringSSL names, so Dart call sites still use `ssl.EVP_PKEY_new()`, `ssl.RAND_bytes()`, etc.
  - `lib/src/third_party/boringssl/ffigen.yaml` is now the source of truth for the BoringSSL entry points bound from Dart. The old `src/symbols.yaml` lookup-table source and generated lookup-table files have been removed.
  - `webcrypto.c` only keeps the remaining native helper for `sizeof(CBB)`, which Dart FFI cannot compute directly.
  - The `EVP_PKEY_free` finalizer no longer needs a C helper; it uses `Native.addressOf(ssl.EVP_PKEY_free)` from the
  generated direct binding.


tool/generate_direct_bindings.dart
```diff
@@ -0,0 +1,255 @@
...
+  const _MethodBinding({
+    required this.name,
+    required this.returnType,
+    required this.params,
+    required this.args,
+    required this.nativeType,
+  });
+}
+
+Future<void> main() async {
```
jonasfj: okay, this is a little special, I guess I'm curious: why not use ffigen, doesn't it support this direct binding mode. I'm pretty sure it's supposed to.

I'm guessing that if we want to do the prefixing thing it might be hard. But perhaps we shouldn't do that? Should we at-least explore not doing that?
It's also possible we could simplify do that by generating C code wrapping the functions, but that also seems like likely to be hard 🙈 

I'm not saying I know what the right thing to do is.

HamdaanAliQuatil: I switched this over to ffigen’s direct binding mode and removed the custom generator.

I kept symbol prefixing for now to avoid silently linking against another BoringSSL revision. ffigen parses the prefixed declarations and renames only the Dart API back to the original names. This also lets us remove the generated wrapper and Windows export manifest.

CI is passing across all platforms.

src/symbols.yaml
```diff
@@ -13,38 +13,17 @@
 # limitations under the License.
 
 ### List of BoringSSL symbols used
-# This file is the source of truth for which BoringSSL symbols can be used from
-# in the Dart code using`dart:ffi`.
+# This file is the source of truth for which prefixed BoringSSL entry points
```
jonasfj: Is this still true. What are we using this file for now?

jonasfj (commented):
Please outline how this works in the PR description

src/webcrypto.c
```diff
@@ -14,15 +14,15 @@
...
 
-WEBCRYPTO_EXPORT void* webcrypto_lookup_symbol(int32_t index) {
-  return _webcrypto_symbol_table[index];
-}
-
 WEBCRYPTO_EXPORT size_t webcrypto_get_CBB_size(void) {
   return sizeof(CBB);
 }
+
+WEBCRYPTO_EXPORT void* webcrypto_get_EVP_PKEY_free_address(void) {
```
jonasfj: Do we still need this hack? Or can we get the address from the ffigen bindings?

HamdaanAliQuatil: I checked ffigen’s current behavior: `symbol-address` is ignored for functions generated as direct `@Native` bindings, so I don’t think we can get the finalizer pointer from the generated direct binding today.

I’d avoid reintroducing lookup just for this. The current helper is small and only exists because `NativeFinalizer` needs a raw native function pointer. I’ll document that explicitly in `webcrypto.c`.

src/symbols.yaml
HamdaanAliQuatil: this file is more of a manual allowlist. 

since the file no longer generates the runtime lookup table, keeping it as a second manual allowlist is probably misleading. 

i think the better cleanup is to remove `src/symbols.yaml` and treat `lib/src/third_party/boringssl/ffigen.yaml` as the actual source of truth for the BoringSSL entry points we bind.

src/webcrypto.c
HamdaanAliQuatil: Update: Liam pointed out `Native.addressOf`, which is the right API here.

I switched this to use `Native.addressOf(ssl.EVP_PKEY_free)` and removed the C helper entirely. So yes, we can get the address from the direct ffigen binding after all 🎉

jonasfj (approved):

PR #281, opened 2026-06-01
docs: make dartdoc examples self-contained

### LINKED ISSUE #101: Make dartdoc examples self-contained.

miachillgood:
Closes #101.

## Summary
- Wrap dartdoc examples that include imports in `main()` functions so they can be copied into runnable Dart files.
- Use `Future<void> main() async` for examples that await webcrypto APIs and `void main()` for the random-byte example.
- Fix a few `keypair`/wording typos in the touched examples while preserving the documented APIs.

## Validation
- `git diff --check`
- Custom script verified all dartdoc `dart` code blocks with imports now include a `main()` function.

### DIFF
--- a/lib/src/webcrypto/webcrypto.aescbc.dart
+++ b/lib/src/webcrypto/webcrypto.aescbc.dart
@@ -56,19 +56,21 @@ final class AesCbcSecretKey {
   /// import 'dart:typed_data' show Uint8List;
   /// import 'package:webcrypto/webcrypto.dart';
   ///
-  /// final rawKey = Uint8List(16);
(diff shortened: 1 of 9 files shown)


jonasfj (approved):

jonasfj:
@miachillgood thanks for the fix, if you want I filed https://github.com/google/webcrypto.dart/issues/283 with more fun :D

Ideally, we can one day make a setup where we run the examples to keep the honest.

PR #290, opened 2026-06-16
Add randomUUID() support

### LINKED ISSUE #187: Consider adding support for randomUUID

Syed-Moiz-Ali:
Fixes #187

Adds \
andomUUID()\ function matching the \Crypto.randomUUID()\ Web API:

- **JS backend**: delegates to \window.crypto.randomUUID()\ via \dart:js_interop\
- **FFI backend**: generates UUID v4 using BoringSSL \RAND_bytes\ with correct version (4) and variant (RFC 4122) bits
- **Stub backend**: throws \UnimplementedError\

The public API is a top-level \String randomUUID()\ function in \package:webcrypto\.

Tests verify:
- UUID format (length, dashes, version nibble, variant nibble, hex characters)
- Uniqueness across 100 generated UUIDs

### DIFF
--- a/lib/src/crypto_subtle.dart
+++ b/lib/src/crypto_subtle.dart
@@ -71,6 +71,9 @@ extension type JSCrypto(JSObject _) implements JSObject {
 
   /// https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
   external JSTypedArray getRandomValues(JSTypedArray array);
+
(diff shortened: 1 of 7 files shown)


HamdaanAliQuatil:
this is planned for [1.1.0](https://github.com/google/webcrypto.dart/milestone/2). putting this on hold as there will be major architectural drifts with native assets and JCA work.

PR #295, opened 2026-06-24
ci: add FFI memory-safety test lane

### LINKED ISSUE #278: FFI safety: add valgrind and memory-pressure testing
Before `1.0.0`, we should add stronger memory-safety testing for the FFI/native path.

The current test suite gives us good functional coverage, but native crypto code also needs protection against:

  - leaks
  - use-after-free
  - invalid ownership transfer
  - lifetime bugs across scopes/finalizers
  - GC-related assumptions that only fail under stress

It may be easy to miss cases where GC, scoped lifetimes, or native ownership expectations interact badly, even if ordinary correctness tests stay green.

## Areas of concern

  The FFI implementation does a lot of explicit lifetime/ownership management, especially in:

  - `lib/src/impl_ffi/impl_ffi.utils.dart`
    - `_PKEYWrapper`
    - `NativeFinalizer`
    - `_Scope`
    - `_SslAllocator`
    - `move(...)`, `defer(...)`, `create(...)`, `dataAsPointer(...)`
  - many FFI call sites under `lib/src/impl_ffi/`
    - AES, RSA, EC, HMAC, HKDF, PBKDF2, digest, random, etc.
  - wrapper/native ownership boundaries in exported/imported key code
  - places where BoringSSL takes ownership and the Dart side intentionally "moves" handles out of scope

## Goal

  Add explicit safety-oriented validation for the native path, not just functional tests.

  ## Proposed work

  ### 1. Valgrind on Linux
  Add at least one CI lane that runs native VM tests under `valgrind` on Linux.

  Initial target:
  - `dart test test/webcrypto_test.dart`

  Later we can decide whether to expand coverage selectively.

  ### 2. Memory-pressure / lifetime stress tests
  Add targeted tests that try to shake out:
  - early frees
  - invalid scope assumptions
  - ownership mistakes
  - finalizer/lifetime issues

 ### 3. Possibly later sanitizer coverage
  If practical, evaluate native sanitizer coverage, especially:

  - ASan for use-after-free / invalid memory access
  - LSan for leaks

  UBSan may also be worth evaluating later, but it is lower priority than memory-access and leak detection for this codebase.

HamdaanAliQuatil:
Closes #278

  This adds a focused memory-safety workload for the native FFI path and runs it under Valgrind on Linux CI.

  The implementation is split into:

  - `test/ffi/valgrind_test.dart`: a VM-only test that skips outside Linux or when Valgrind is unavailable, then launches
  the standalone target under Valgrind.
  - `test/ffi/valgrind_target.dart`: a standalone Dart program without `package:test`, keeping the monitored process
  focused on native crypto operations and memory pressure.

  What this covers:

  - AES-GCM, HMAC, ECDH, and RSA-OAEP operations through the public API.
  - Key export/import ownership paths for raw, JWK, PKCS8, and SPKI formats.
  - Failed native imports followed by valid operations, exercising cleanup-after-error paths.
  - Repeated short-lived allocations between native operations to put pressure on Dart object lifetimes and finalizers.

  The CI job uses the normal `dart test` path because it supports native build hooks today. Sanitizer support for hook-
  built code remains blocked upstream, so this starts with Valgrind Memcheck.

  Valgrind reports definite and possible leaks but only fails on definite leaks. Possible leaks can include Dart VM
  runtime noise, so they remain visible without making the initial CI lane unnecessarily brittle.


test/ffi_memory_safety_test.dart
```diff
@@ -0,0 +1,261 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+@TestOn('vm')
+@Timeout(Duration(minutes: 3))
+library;
```
jonasfj: May I suggest we re-organize this a bit.


```
test/
  ffi/
    valgrind_test.dart     // void main() => test( => process.runSync(valgrind...
    valgrind_target.dart   // void main() {...
```

So that `valgrind_test.dart` is annotated `@TestOn('vm')` and it has:
`if(!linux) { markTestSkipped('only test on linux'); return; }`

Probably, we could even run `Process.runSync(which valgrind` and skip the test if `valgrind` is not installed.

`valgrind_target.dart` is just a `main` with `print("#symmetric operations survive memory pressure")` doesn't use `package:test`, it can still throw `AssertionError` or use `package:checks` to make assertions. But it's really not about testing logic, it's about calling methods and doing memory pressure things and letting valgrind have some behavior that it can inspect.

HamdaanAliQuatil: Done. I split it into a small package:test controller and a standalone target without test-runner dependencies. The controller skips outside Linux or when Valgrind is unavailable, and Valgrind now only monitors the crypto workload and memory-pressure operations.

test/ffi/valgrind_test.dart
```diff
@@ -0,0 +1,55 @@
...
+
+@TestOn('vm')
+@Timeout(Duration(minutes: 3))
+library;
+
+import 'dart:io';
+
+import 'package:test/test.dart';
+
+void main() {
```
jonasfj, line 25: Please consider adding a documentation comment explaining this file and how it's associated with valgrind target, and what purpose is.

.github/workflows/test.yml
```diff
@@ -58,6 +58,20 @@ jobs:
...
+    name: Linux FFI memory safety
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dart-lang/setup-dart@v1
+      - name: Install Valgrind
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y valgrind
```
jonasfj, line 62: Do we need a separate GitHub actions job at all?

Is this super slow. If not let's just add it to the existing tests.

We just have to add this install valgrind line, right?

test/ffi/valgrind_target.dart
```diff
@@ -0,0 +1,319 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:async';
+import 'dart:typed_data';
+
+import 'package:webcrypto/webcrypto.dart';
+
+Future<void> main() async {
```
jonasfj, line 23: Please consider a documentation comment, explaining where this file is used, why it is the way it is..

What we hope to gain from this madness :)

jonasfj (approved):
I think we should consider avoiding the extra GitHub actions job if not strictly necessary

.github/workflows/test.yml
HamdaanAliQuatil: Agreed. I removed the separate job and folded Valgrind into the existing linux-dart job

PR #296, opened 2026-06-25
feat(impl_jni): add JCA backend skeleton and digest path

mfazrinizar:
## Summary

This PR adds the first Android JCA exploration slice on top of `android-jca-branch`.

It includes:
- an experimental `lib/src/impl_jni` backend skeleton following `WebCryptoImpl`
- JCA `MessageDigest` support for SHA-1, SHA-256, SHA-384, and SHA-512
- scoped JNI object release for transient Java objects
- a `JByteArray.copyToDartBytes()` helper so digest output is copied into Dart-owned memory
- temporary native import wiring to use the JNI backend on this exploration branch
- focused desktop JNI digest tests
- an Android example integration smoke test proving the public API uses the JNI digest path

Generated JNI bindings are intentionally not part of this PR. They were split out and landed separately in #297.

Non-digest primitives are still stubbed. This is intentional for the first slice so the branch establishes backend shape, digest behavior, and testing flow before adding HMAC/AES-GCM/etc.

## Testing

```sh
dart analyze
dart run jni:setup
dart test test/impl_jni_digest_test.dart
```

Android, inside example app directory:
```sh
flutter test integration_test/jni_digest_test.dart -d emulator-name
```

### DIFF
--- a/example/integration_test/jni_digest_test.dart
+++ b/example/integration_test/jni_digest_test.dart
@@ -1,0 +1,22 @@
+import 'dart:convert';
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+import 'package:webcrypto/src/impl_jni/impl_jni.dart' as jni_impl;
+import 'package:webcrypto/webcrypto.dart';
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  testWidgets('JCA digest works through the public API', (_) async {
+    final data = utf8.encode('hello-world');
+    final digest = await Hash.sha256.digestBytes(data);
+    final jcaDigest = await jni_impl.webCryptImpl.sha256.digestBytes(data);
+
+    expect(digest, jcaDigest);
+    expect(
+      base64Encode(digest),
+      'r6J7RNQ7Aqn+pB0TztwuQBbPz4fF2/mQ5ZNmmqjOKG0=',
+    );
+  });
+}
--- a/example/integration_test/jni_digest_test.dart
+++ b/example/integration_test/jni_digest_test.dart
@@ -2,18 +2,15 @@ import 'dart:convert';
 
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
-import 'package:webcrypto/src/impl_jni/impl_jni.dart' as jni_impl;
 import 'package:webcrypto/webcrypto.dart';
 
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
 
-  testWidgets('JCA digest works through the public API', (_) async {
+  testWidgets('public API SHA-256 digest matches known vector', (_) async {
     final data = utf8.encode('hello-world');
     final digest = await Hash.sha256.digestBytes(data);
-    final jcaDigest = await jni_impl.webCryptImpl.sha256.digestBytes(data);
 
-    expect(digest, jcaDigest);
     expect(
       base64Encode(digest),
       'r6J7RNQ7Aqn+pB0TztwuQBbPz4fF2/mQ5ZNmmqjOKG0=',
--- a/lib/src/impl_jni/impl_jni.digest.dart
+++ b/lib/src/impl_jni/impl_jni.digest.dart
@@ -42,13 +42,14 @@ final class _HashImpl implements HashImpl {
 
   @override
   Future<Uint8List> digestStream(Stream<List<int>> data) async {
-    final algorithm = jni.JString.fromString(_jcaName);
-    final digest = MessageDigest.getInstance(algorithm);
-    algorithm.release();
-
-    if (digest == null) {
-      throw operationError('JCA MessageDigest($_jcaName) is unavailable');
-    }
+    final digest = jni.using((arena) {
+      final algorithm = jni.JString.fromString(_jcaName)..releasedBy(arena);
+      final digest = MessageDigest.getInstance(algorithm);
+      if (digest == null) {
+        throw operationError('JCA MessageDigest($_jcaName) is unavailable');
+      }
+      return digest;
+    });
 
     try {
       await for (final chunk in data) {
--- a/lib/src/impl_jni/impl_jni.digest.dart
+++ b/lib/src/impl_jni/impl_jni.digest.dart
@@ -1,0 +1,74 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+part of 'impl_jni.dart';
+
+final class _HashImpl implements HashImpl {
+  const _HashImpl(this._jcaName);
+
+  final String _jcaName;
+
+  @override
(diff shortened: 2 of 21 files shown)


tool/generate_jni_bindings.dart
```diff
@@ -0,0 +1,58 @@
...
+/// The initial binding set covers the P0 primitives from the GSoC proposal:
+/// digest, HMAC, and AES-GCM. Later milestones can extend [classes] as more
+/// JCA primitives are ported behind `WebCryptoImpl`.
+library;
+
+import 'dart:io';
+
+import 'package:jnigen/jnigen.dart';
+
+Future<void> main() async {
```
jonasfj: Could we just use: a `jnigen.yaml` config file, see:
https://pub.dev/packages/jnigen#yaml-configuration-reference

Add it to:
https://github.com/google/webcrypto.dart/blob/master/tool/update-bindings.sh

Put the generated bindings in:
```
lib/
  src/
    third_party/
      jca/
        jnigen.yaml
        generated_bindings.dart
```

Or something like that

lib/src/impl_jni/impl_jni.digest.dart
```diff
@@ -0,0 +1,85 @@
...
+    }
+  }
+}
+
+/// Copies a JVM byte array into Dart-owned memory.
+///
+/// `JByteArray.getRange` returns a typed list backed by a native buffer whose
+/// lifetime is finalizer-driven. Copy once more before releasing the Java array
+/// so webcrypto callers receive a normal Dart-managed [Uint8List].
+Uint8List _copyJByteArray(jni.JByteArray array) {
```
jonasfj: consider putting this in a util.dart library or similar...

consider using an extension method on `JByteArray`, so you can do `myArray.copyToDart()` or something like that :D

lib/src/impl_native/impl_native.dart
```diff
@@ -0,0 +1,35 @@
...
+
+const _requestedBackend = String.fromEnvironment('webcrypto.impl');
+
+const WebCryptoImpl webCryptImpl = _requestedBackend == 'jni'
+    ? jni.webCryptImpl
+    : ffi.webCryptImpl;
+
+@visibleForTesting
+WebCryptoImpl selectNativeImplForTesting(String requestedBackend) =>
+    requestedBackend == 'jni' ? jni.webCryptImpl : ffi.webCryptImpl;
```
jonasfj: what is this good for? probably don't need it :D

mfazrinizar: This was only added to make the temporary selector easy to test because I wasn’t really sure the `-D` path would be reliable through package:test. But agreed, it is not important right now. I’ll remove it and keep the tests focused on direct JNI/JCA smoke coverage.

lib/src/impl_jni/impl_jni.digest.dart
mfazrinizar: Makes sense. I’ll move this into a small JNI util library and expose it as an extension method. I think probably something like `copyToDartBytes()`, so that the digest implementation stays focused on the crypto flow.

tool/generate_jni_bindings.dart
mfazrinizar: Agreed. We should use the config-driven generation approach. I’ll switch this to a config-based setup matching the existing bindings flow: add `lib/src/third_party/jca/jnigen.yaml`, wire it into `tool/update-bindings.sh`, and put the generated output under `lib/src/third_party/jca/generated_bindings.dart`. In this case, I’ll split the generated bindings into a separate PR first so this implementation PR only contains the backend/test code.

mfazrinizar:
Rebased this on top of the merged generated-bindings PR (#297). The diff now only contains the code JNI backend skeleton, digest implementation, and tests. I also reran the focused desktop digest test and Android 16 emulator smoke test successfully.

test/impl_jni_digest_test.dart
```diff
@@ -0,0 +1,74 @@
...
+
+    final helperDir = Directory('build/jni_libs');
+    final helperName = Platform.isWindows
+        ? 'dartjni.dll'
+        : Platform.isMacOS
+        ? 'libdartjni.dylib'
+        : 'libdartjni.so';
+    if (!File.fromUri(helperDir.uri.resolve(helperName)).existsSync()) {
+      markTestSkipped('Run `dart run jni:setup` before desktop JNI tests.');
+      return;
```
HamdaanAliQuatil: `markTestSkipped` inside `setUpAll` does not skip the tests themselves. I confirmed that without `build/jni_libs`, the setup is marked skipped but both tests still execute and fail with `HelperNotFoundError`.

please either determine availability before registering the tests and use `skip:`, or fail early with a clear prerequisite. the PR’s documented test command should also mention `dart run jni:setup` if the helper is required.

the same ambiguity exists in `jni_test_setup_io.dart:32`: silently returning does not make the JNI-backed suite usable.

lib/src/impl_jni/impl_jni.digest.dart
```diff
@@ -0,0 +1,74 @@
...
+      result.releasedBy(arena);
+
+      return result.copyToDartBytes();
+    });
+  }
+
+  @override
+  Future<Uint8List> digestStream(Stream<List<int>> data) async {
+    final algorithm = jni.JString.fromString(_jcaName);
+    final digest = MessageDigest.getInstance(algorithm);
```
HamdaanAliQuatil: nit: If `MessageDigest.getInstance(algorithm)` throws, `algorithm.release()` is never reached. could this creation be wrapped in `try/finally` or a small `jni.using` scope, consistent with `digestBytes`?

example/integration_test/jni_digest_test.dart
```diff
@@ -0,0 +1,22 @@
...
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+import 'package:webcrypto/src/impl_jni/impl_jni.dart' as jni_impl;
+import 'package:webcrypto/webcrypto.dart';
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  testWidgets('JCA digest works through the public API', (_) async {
```
HamdaanAliQuatil: nit: this verifies that the public api returns the correct digest, but it does not prove that jni was selected because boringssl produces the same result.

please we rename the test accordingly. the conditional import already demonstrates the routing.

HamdaanAliQuatil (changes requested):

test/impl_jni_digest_test.dart
mfazrinizar: Fixed. The focused desktop JNI digest test now determines helper availability before registering test behavior and uses `skip:` instead of `markTestSkipped()` from `setUpAll`.

The shared desktop JNI setup now fails early with a clear `dart run jni:setup` prerequisite when a JNI-backed suite requires the helper, instead of silently returning. I have also updated the PR testing notes to include `dart run jni:setup`.

lib/src/impl_jni/impl_jni.digest.dart
mfazrinizar: Fixed. `digestStream` now creates the algorithm string inside a `jni.using` scope before calling `MessageDigest.getInstance`, so the transient `JString` is released even if creation throws.

example/integration_test/jni_digest_test.dart
mfazrinizar: Fixed. I renamed and simplified the Android smoke test so it only claims public API SHA-256 correctness. It no longer claims to prove backend selection.

test/impl_jni_digest_test.dart
mfazrinizar: Fixed. The focused desktop JNI digest test now determines helper availability before registering test behavior and uses `skip:` instead of `markTestSkipped()` from `setUpAll`. And the shared desktop JNI setup, it now fails early with a clear `dart run jni:setup` prerequisite when a JNI-backed suite requires the helper, instead of silently returning. I have also updated the PR testing notes to include `dart run jni:setup`.

HamdaanAliQuatil (approved):

PR #297, opened 2026-06-25
chore(jca): add generated JNI bindings

mfazrinizar:
## Summary

This PR lands the JNI/JCA generated bindings separately so later Android JCA backend code is easier to review (as per suggestion https://github.com/google/webcrypto.dart/pull/296#discussion_r3474681602).

It includes two commits:

- add `jnigen`/`jni` dependencies, `lib/src/third_party/jca/jnigen.yaml`, and `tool/update-bindings.sh` wiring
- add the generated JNI bindings under `lib/src/third_party/jca/generated_bindings.dart`

The generated bindings currently cover the Java/JCA classes needed for the initial Android JCA exploration:
- `java.security.MessageDigest`
- `java.security.SecureRandom`
- `javax.crypto.Cipher`
- `javax.crypto.KeyGenerator`
- `javax.crypto.Mac`
- `javax.crypto.spec.GCMParameterSpec`
- `javax.crypto.spec.SecretKeySpec`

## Testing

This intentionally does not include  `impl_jni` backend code. That will stay in the follow-up implementation PR so generated code does not pollute the review. `flutter build apk` is needed first so JNIgen can collect the Android Gradle classpath from the example app. 

Inside example app directory:
```sh
flutter build apk
dart run jnigen --config=lib/src/third_party/jca/jnigen.yaml
dart analyze
```


HamdaanAliQuatil (commented):
Thanks for splitting this out. This is the right direction: config/script changes in one commit, generated bindings in a separate commit.

I think this PR should stay focused on reproducible generated bindings. A few things I would like clarified/fixed before landing:

1. `tool/update-bindings.sh` now runs `jnigen` unconditionally. I agree that the JCA bindings need to be generated and checked in, but this changes the existing binding-regeneration workflow because anyone running the script now also needs the JNI/Android generation environment.

Is that intended for `android-jca-branch`? If yes, please document the required setup/API level and the exact command. If not, it may be cleaner to put the JCA/JNI binding generation behind a separate script for now.

2. The `jnigen.yaml` uses `android_sdk_config.add_gradle_deps: true` with `android_example: ../../../../example`. Does running the generator mutate any example/Gradle files? Please confirm the exact regenerate command and that `git status --short` is clean after regeneration.

3. Please document the Android SDK/API level used for generation. Since this is generated from platform APIs, we need the generated output to be reproducible and not depend on whichever SDK/JDK happens to be installed locally.

I do not think we need to line-review the generated Dart file itself once the generation inputs are reproducible.

mfazrinizar:
> Thanks for splitting this out. This is the right direction: config/script changes in one commit, generated bindings in a separate commit.
> 
> I think this PR should stay focused on reproducible generated bindings. A few things I would like clarified/fixed before landing:
> 
> 1. `tool/update-bindings.sh` now runs `jnigen` unconditionally. I agree that the JCA bindings need to be generated and checked in, but this changes the existing binding-regeneration workflow because anyone running the script now also needs the JNI/Android generation environment.
> 
> Is that intended for `android-jca-branch`? If yes, please document the required setup/API level and the exact command. If not, it may be cleaner to put the JCA/JNI binding generation behind a separate script for now.
> 
> 2. The `jnigen.yaml` uses `android_sdk_config.add_gradle_deps: true` with `android_example: ../../../../example`. Does running the generator mutate any example/Gradle files? Please confirm the exact regenerate command and that `git status --short` is clean after regeneration.
> 3. Please document the Android SDK/API level used for generation. Since this is generated from platform APIs, we need the generated output to be reproducible and not depend on whichever SDK/JDK happens to be installed locally.
> 
> I do not think we need to line-review the generated Dart file itself once the generation inputs are reproducible.

Thanks for the comment. I changed this so the existing `tool/update-bindings.sh` no longer runs JNIgen unconditionally. That keeps the existing BoringSSL binding regeneration workflow from requiring Android/JNI setup.

JCA/JNI binding generation is now in a dedicated script: `tool/update-jca-bindings.sh`

I also changed `jnigen.yaml` to avoid Gradle classpath collection. Instead of `add_gradle_deps: true` + `android_example`, it now pins Android SDK platform stubs directly: `android_sdk_config.versions: ['36']`

So the generated bindings are based on Android API 36’s `android.jar`, not whichever SDK/Gradle classpath happens to be local.

Also, added a small README next to the JCA JNI binding config documenting the regeneration command, Android SDK Platform 36 requirement, `ANDROID_SDK_ROOT`, and why the API level is pinned for reproducibility.

lib/src/third_party/jca/README.md
```diff
@@ -0,0 +1,26 @@
+# JCA JNI Bindings
+
+These bindings are generated with JNIgen for the Android JCA backend
+exploration.
+
+To regenerate them, run from the package root:
+
+```sh
+tool/update-jca-bindings.sh
```
HamdaanAliQuatil:
```suggestion
bash tool/update-jca-bindings.sh
```

HamdaanAliQuatil (approved):

PR #298, opened 2026-06-29
feat(impl_jni): add JCA HMAC implementation

mfazrinizar:
## Summary

Adds the JNI/JCA HMAC implementation on top of the existing Android JCA backend skeleton.

This PR implements:
- HMAC raw key import/export
- HMAC JWK import/export
- HMAC key generation using JCA `SecureRandom`
- HMAC signing and verification through JCA `Mac`
- byte and stream signing/verification
- non-byte-aligned key length handling
- focused desktop JNI tests and Android smoke test

## Testing

- `dart run jni:setup`

Verified with:
- `dart analyze`
- `dart test test/impl_jni_hmac_test.dart`
- `dart test test/webcrypto_test.dart -p vm -n HMAC`
- `dart test test/webcrypto_test.dart -p vm -n Random`
- `flutter test integration_test/jni_hmac_test.dart -d emulator-name`

### DIFF
--- a/lib/src/impl_jni/impl_jni.hmac.dart
+++ b/lib/src/impl_jni/impl_jni.hmac.dart
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -14,8 +14,6 @@
 
 part of 'impl_jni.dart';
 
-const _hmacStreamBufferLength = 4096;
-
 _HashImpl _hmacHashFromHash(HashImpl hash) {
   if (hash is _HashImpl) {
     return hash;
@@ -30,7 +28,7 @@ extension _HmacHashMetadata on _HashImpl {
       'SHA-256' => 'HmacSHA256',
       'SHA-384' => 'HmacSHA384',
       'SHA-512' => 'HmacSHA512',
-      _ => throw AssertionError('Unsupported hash algorithm: $_jcaName'),
+      _ => throw AssertionError('Unknown hash algorithm: $_jcaName'),
     };
   }
 
@@ -40,7 +38,7 @@ extension _HmacHashMetadata on _HashImpl {
       'SHA-256' => 'HS256',
       'SHA-384' => 'HS384',
       'SHA-512' => 'HS512',
-      _ => throw AssertionError('Unsupported hash algorithm: $_jcaName'),
+      _ => throw AssertionError('Unknown hash algorithm: $_jcaName'),
     };
   }
 
@@ -50,7 +48,7 @@ extension _HmacHashMetadata on _HashImpl {
       'SHA-256' => 256,
       'SHA-384' => 384,
       'SHA-512' => 512,
-      _ => throw AssertionError('Unsupported hash algorithm: $_jcaName'),
+      _ => throw AssertionError('Unknown hash algorithm: $_jcaName'),
     };
   }
 }
@@ -116,7 +114,7 @@ final class _HmacSecretKeyImpl implements HmacSecretKeyImpl {
   Future<Uint8List> signBytes(List<int> data) async {
     return jni.using((arena) {
       final mac = _createMac(arena);
-      final input = arena.copyToJByteArray(data);
+      final input = arena.copyToJByteArray(_asUint8List(data));
       return _copyMacResult(arena, mac.doFinal$2(input));
     });
   }
@@ -126,7 +124,7 @@ final class _HmacSecretKeyImpl implements HmacSecretKeyImpl {
     final arena = jni.Arena();
     try {
       final mac = _createMac(arena);
-      final buffer = jni.JByteArray(_hmacStreamBufferLength)..releasedBy(arena);
+      final buffer = jni.JByteArray(_defaultChunkSize)..releasedBy(arena);
       await for (final chunk in data) {
         _updateMacWithChunk(mac, buffer, chunk);
       }
@@ -175,13 +173,14 @@ final class _HmacSecretKeyImpl implements HmacSecretKeyImpl {
   }
 
   void _updateMacWithChunk(Mac mac, jni.JByteArray buffer, List<int> chunk) {
+    final bytes = _asUint8List(chunk);
     var offset = 0;
-    while (offset < chunk.length) {
-      final remaining = chunk.length - offset;
-      final length = remaining < _hmacStreamBufferLength
+    while (offset < bytes.length) {
+      final remaining = bytes.length - offset;
+      final length = remaining < _defaultChunkSize
           ? remaining
-          : _hmacStreamBufferLength;
-      buffer.setRange(0, length, chunk, offset);
+          : _defaultChunkSize;
+      buffer.setRange(0, length, bytes, offset);
       mac.update$2(buffer, 0, length);
       offset += length;
     }
@@ -198,7 +197,7 @@ final class _HmacSecretKeyImpl implements HmacSecretKeyImpl {
   bool _verifySignature(Uint8List computedMac, List<int> suppliedSignature) {
     return jni.using((arena) {
       final computed = arena.copyToJByteArray(computedMac);
-      final supplied = arena.copyToJByteArray(suppliedSignature);
+      final supplied = arena.copyToJByteArray(_asUint8List(suppliedSignature));
       return MessageDigest.isEqual(computed, supplied);
     });
   }
--- a/lib/src/impl_jni/impl_jni.hmac.dart
+++ b/lib/src/impl_jni/impl_jni.hmac.dart
@@ -14,6 +14,8 @@
 
 part of 'impl_jni.dart';
 
+const _hmacStreamBufferLength = 4096;
+
 _HashImpl _hmacHashFromHash(HashImpl hash) {
   if (hash is _HashImpl) {
     return hash;
@@ -28,7 +30,7 @@ extension _HmacHashMetadata on _HashImpl {
       'SHA-256' => 'HmacSHA256',
       'SHA-384' => 'HmacSHA384',
       'SHA-512' => 'HmacSHA512',
-      _ => throw StateError('Unsupported hash algorithm: $_jcaName'),
+      _ => throw AssertionError('Unsupported hash algorithm: $_jcaName'),
     };
   }
 
@@ -38,7 +40,7 @@ extension _HmacHashMetadata on _HashImpl {
       'SHA-256' => 'HS256',
       'SHA-384' => 'HS384',
       'SHA-512' => 'HS512',
-      _ => throw StateError('Unsupported hash algorithm: $_jcaName'),
+      _ => throw AssertionError('Unsupported hash algorithm: $_jcaName'),
     };
   }
 
@@ -48,7 +50,7 @@ extension _HmacHashMetadata on _HashImpl {
       'SHA-256' => 256,
       'SHA-384' => 384,
       'SHA-512' => 512,
-      _ => throw StateError('Unsupported hash algorithm: $_jcaName'),
+      _ => throw AssertionError('Unsupported hash algorithm: $_jcaName'),
     };
   }
 }
@@ -112,56 +114,37 @@ final class _HmacSecretKeyImpl implements HmacSecretKeyImpl {
 
   @override
   Future<Uint8List> signBytes(List<int> data) async {
-    final mac = _createMac();
-    try {
-      final result = jni.using((arena) {
-        final input = jni.JByteArray.from(data)..releasedBy(arena);
-        final result = mac.doFinal$2(input);
-        if (result == null) {
-          throw operationError('JCA Mac(${_hash._hmacJcaName}) returned null');
-        }
-        result.releasedBy(arena);
-        return result.copyToDartBytes();
-      });
-      return result;
-    } finally {
-      mac.release();
-    }
+    return jni.using((arena) {
+      final mac = _createMac(arena);
+      final input = arena.copyToJByteArray(data);
+      return _copyMacResult(arena, mac.doFinal$2(input));
+    });
   }
 
   @override
   Future<Uint8List> signStream(Stream<List<int>> data) async {
-    final mac = _createMac();
+    final arena = jni.Arena();
     try {
+      final mac = _createMac(arena);
+      final buffer = jni.JByteArray(_hmacStreamBufferLength)..releasedBy(arena);
       await for (final chunk in data) {
-        jni.using((arena) {
-          final input = jni.JByteArray.from(chunk)..releasedBy(arena);
-          mac.update$1(input);
-        });
+        _updateMacWithChunk(mac, buffer, chunk);
       }
 
-      final result = mac.doFinal();
-      if (result == null) {
-        throw operationError('JCA Mac(${_hash._hmacJcaName}) returned null');
-      }
-      try {
-        return result.copyToDartBytes();
-      } finally {
-        result.release();
-      }
+      return _copyMacResult(arena, mac.doFinal());
     } finally {
-      mac.release();
+      arena.releaseAll();
     }
   }
 
   @override
   Future<bool> verifyBytes(List<int> signature, List<int> data) async {
-    return _constantTimeEquals(signature, await signBytes(data));
+    return _verifySignature(await signBytes(data), signature);
   }
 
   @override
   Future<bool> verifyStream(List<int> signature, Stream<List<int>> data) async {
-    return _constantTimeEquals(signature, await signStream(data));
+    return _verifySignature(await signStream(data), signature);
   }
 
   @override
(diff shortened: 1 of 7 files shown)


lib/src/impl_jni/impl_jni.hmac.dart
```diff
@@ -14,6 +14,45 @@
...
+}
+
+extension _HmacHashMetadata on _HashImpl {
+  String get _hmacJcaName {
+    return switch (_jcaName) {
+      'SHA-1' => 'HmacSHA1',
+      'SHA-256' => 'HmacSHA256',
+      'SHA-384' => 'HmacSHA384',
+      'SHA-512' => 'HmacSHA512',
+      _ => throw StateError('Unsupported hash algorithm: $_jcaName'),
```
jonasfj: look at what we do in FFI

IMO, AssertionError or UnsupportedError is more appropriate

lib/src/impl_jni/impl_jni.hmac.dart
```diff
@@ -22,21 +61,153 @@ final class _StaticHmacSecretKeyImpl implements StaticHmacSecretKeyImpl {
...
+        if (result == null) {
+          throw operationError('JCA Mac(${_hash._hmacJcaName}) returned null');
+        }
+        result.releasedBy(arena);
+        return result.copyToDartBytes();
+      });
+      return result;
+    } finally {
+      mac.release();
+    }
```
jonasfj: couldn't this be avoided if you pass the arena into _createMac ?

lib/src/impl_jni/impl_jni.hmac.dart
```diff
@@ -22,21 +61,153 @@ final class _StaticHmacSecretKeyImpl implements StaticHmacSecretKeyImpl {
...
+
+  final Uint8List _keyData;
+  final _HashImpl _hash;
+
+  @override
+  Future<Uint8List> signBytes(List<int> data) async {
+    final mac = _createMac();
+    try {
+      final result = jni.using((arena) {
+        final input = jni.JByteArray.from(data)..releasedBy(arena);
```
jonasfj: if you have to do this often through out the code, maybe an extension method on arena that let's you do `arena.copyToJByteArray(Uint8List data)` could be neat.. Maybe there is a better method name too.

lib/src/impl_jni/impl_jni.hmac.dart
```diff
@@ -22,21 +61,153 @@ final class _StaticHmacSecretKeyImpl implements StaticHmacSecretKeyImpl {
...
+        final result = mac.doFinal$2(input);
+        if (result == null) {
+          throw operationError('JCA Mac(${_hash._hmacJcaName}) returned null');
```
jonasfj: Find out when and why this happens, and if it can happen.

If this can't happen then this is an internal AssertionError.

If this can happen then it failed for a reason, right?

lib/src/impl_jni/impl_jni.hmac.dart
```diff
@@ -22,21 +61,153 @@ final class _StaticHmacSecretKeyImpl implements StaticHmacSecretKeyImpl {
...
+  final Uint8List _keyData;
+  final _HashImpl _hash;
+
+  @override
+  Future<Uint8List> signBytes(List<int> data) async {
+    final mac = _createMac();
+    try {
+      final result = jni.using((arena) {
+        final input = jni.JByteArray.from(data)..releasedBy(arena);
+        final result = mac.doFinal$2(input);
```
jonasfj: this is a bit crazy, but one could do:
```dart
extension<T1> on JByteArray Function(T1) {
  JByteArray invoke(jni.Arena arena, T1 arg1) {
    final r = this(arg1);
    if (r != null) {
      // TODO: Check if "$this" prints to something sensible
      throw AssertionError('$this returned null which is impossible');
    }
    r.releasedBy(arena)
    return r;
  }
}
```

Then one can simply write:
```dart
final result = mac.doFinal$2.invoke(arena, input);
```


I did something similar here:
https://github.com/google/webcrypto.dart/blob/c59eb72c8c51af690231bb292b1bde0f720d1b93/lib/src/impl_ffi/impl_ffi.utils.dart#L55-L83

You can debate whether it's brilliant or insane, what you prefer is up to personal preferences.

lib/src/impl_jni/impl_jni.hmac.dart
```diff
@@ -22,21 +61,153 @@ final class _StaticHmacSecretKeyImpl implements StaticHmacSecretKeyImpl {
...
+  _HmacSecretKeyImpl(this._keyData, this._hash);
+
+  final Uint8List _keyData;
+  final _HashImpl _hash;
+
+  @override
+  Future<Uint8List> signBytes(List<int> data) async {
+    final mac = _createMac();
+    try {
+      final result = jni.using((arena) {
```
jonasfj:
```suggestion
      return jni.using((arena) {
```

jonasfj: Consider `import package:jni/jni.dart show using`

you'll be "using" it a lot, right?

When you do `using((arena) {` consider `using((a) {`

If you use `a` consistently then the short hand can be a decent idea.
But only if you consistently do it everywhere.

lib/src/impl_jni/impl_jni.hmac.dart
```diff
@@ -22,21 +61,153 @@ final class _StaticHmacSecretKeyImpl implements StaticHmacSecretKeyImpl {
...
+    }
+  }
+
+  @override
+  Future<Uint8List> signStream(Stream<List<int>> data) async {
+    final mac = _createMac();
+    try {
+      await for (final chunk in data) {
+        jni.using((arena) {
+          final input = jni.JByteArray.from(chunk)..releasedBy(arena);
```
jonasfj: You are making an allocation for every chunk.

At minimum leave a TODO to fix this.

You can just allocate 4096 bytes of JByteArray upfront... Then copy into it, and use a mac.update where you give it an int saying length from JByteArray to consume for the final thing, or something like that.

It's much more complicated, but fewer times you cross the JNI barrier the better, and the few allocations the better.

This is a buffer you can control.. I'm just guessing that 4096 is a good chunk size.

lib/src/impl_jni/impl_jni.hmac.dart
```diff
@@ -22,21 +61,153 @@ final class _StaticHmacSecretKeyImpl implements StaticHmacSecretKeyImpl {
...
+      } catch (_) {
+        mac.release();
+        rethrow;
+      }
+      return mac;
+    });
+  }
+}
+
+bool _constantTimeEquals(List<int> a, List<int> b) {
```
jonasfj:
```suggestion
bool _constantTimeEquals(Uint8List a, Uint8List b) {
```

never operate on List<int> horrible perf

jonasfj: check if we already have an implementation you can reuse

lib/src/impl_jni/impl_jni.utils.dart
```diff
@@ -14,6 +14,49 @@
 
 part of 'impl_jni.dart';
 
+void _checkData(bool condition, String message) {
+  if (!condition) {
+    throw FormatException(message);
+  }
+}
+
+Uint8List _asUint8ListZeroedToBitLength(List<int> data, [int? lengthInBits]) {
```
jonasfj, line 29: single line documentation comment, be concise

lib/src/impl_jni/impl_jni.utils.dart
```diff
@@ -14,6 +14,49 @@
...
+    return base64Url.decode(padded);
+  } on FormatException {
+    throw FormatException(
+      'JWK property "$prop" is not url-safe base64 without padding',
+      unpadded,
+    );
+  }
+}
+
+String _jwkEncodeBase64UrlNoPadding(List<int> data) {
```
jonasfj: Uint8List if possible

lib/src/impl_jni/impl_jni.utils.dart
```diff
@@ -26,3 +69,22 @@ extension _JByteArrayCopy on jni.JByteArray {
...
+}
+
+void _fillRandomBytes(TypedData destination) {
+  final output = destination.buffer.asUint8List(
+    destination.offsetInBytes,
+    destination.lengthInBytes,
+  );
+  jni.using((arena) {
+    final random = SecureRandom()..releasedBy(arena);
+    final bytes = jni.JByteArray(output.length)..releasedBy(arena);
```
jonasfj: consider the size of the buffers here.. if the destination is small fine, but if the destination is > 4096, consider allocating a single 4096 buffer and then looping through and copying bytes over

lib/src/impl_jni/impl_jni.hmac.dart
HamdaanAliQuatil: this is also a security concern, not only a list<int> performance issue.

verifybytes and verifystream promise fixed-time comparison, but a handwritten dart loop is not a reliable constant-time primitive under jit/aot optimization. the generated bindings already expose messagedigest.isequal, whose android contract says timing depends only on the first argument’s length.

please use `messagedigest.isequal`, passing the computed mac first and the supplied signature second, and remove `_constanttimeequals`.

HamdaanAliQuatil (changes requested):

lib/src/impl_jni/impl_jni.hmac.dart
mfazrinizar: Switched impossible internal mapping cases to AssertionError.

lib/src/impl_jni/impl_jni.hmac.dart
mfazrinizar: _createMac now takes the arena and registers transient JNI objects there.

lib/src/impl_jni/impl_jni.hmac.dart
mfazrinizar: Added arena.copyToJByteArray(...) and used it for key/input copies.

lib/src/impl_jni/impl_jni.hmac.dart
mfazrinizar: null Java wrappers are treated as internal AssertionError cases.

lib/src/impl_jni/impl_jni.hmac.dart
mfazrinizar: Kept this simpler with explicit helpers for now instead of a generic invoke extension.

lib/src/impl_jni/impl_jni.hmac.dart
mfazrinizar: Applied..

mfazrinizar: Kept jni.using/jni.Arena explicit for consistency with current impl_jni style.

lib/src/impl_jni/impl_jni.hmac.dart
mfazrinizar: signStream now reuses one 4096-byte JByteArray and calls mac.update$2(buffer, 0, length). Let me know if this is OK.

lib/src/impl_jni/impl_jni.hmac.dart
mfazrinizar: Removed the Dart comparison loop and now use MessageDigest.isEqual.

lib/src/impl_jni/impl_jni.utils.dart
mfazrinizar: Shhortened the copyToDartBytes doc comment.

lib/src/impl_jni/impl_jni.hmac.dart
mfazrinizar: Reused JCA MessageDigest.isEqual.

lib/src/impl_jni/impl_jni.utils.dart
mfazrinizar: CHanged the JWK encode helper to take Uint8List.

lib/src/impl_jni/impl_jni.utils.dart
mfazrinizar: _fillRandomBytes now uses a max 4096-byte Java buffer and loops for larger outputs.

lib/src/impl_jni/impl_jni.hmac.dart
mfazrinizar: Removed _constantTimeEquals andverify now uses MessageDigest.isEqual(computedMac, suppliedSignature), with computed MAC passed first. Added a mismatched-length signature test too.

lib/src/impl_jni/impl_jni.hmac.dart
jonasfj: mostly suggesting that maybe it's worth considering at some point.

But yes, probably better done in a separate PR.

lib/src/impl_jni/impl_jni.utils.dart
```diff
@@ -14,15 +14,91 @@
...
+  }
+
+  jni.using((arena) {
+    final random = SecureRandom()..releasedBy(arena);
+    final bufferLength = output.length < 4096 ? output.length : 4096;
+    final bytes = jni.JByteArray(bufferLength)..releasedBy(arena);
+
+    var offset = 0;
+    while (offset < output.length) {
+      random.nextBytes(bytes);
```
jonasfj: shouldn't request more bytes than we need... figure out chunkLength first,

I don't think this is likely to have practical effects, but worth doing.

lib/src/impl_jni/impl_jni.utils.dart
```diff
@@ -14,15 +14,91 @@
...
+      final remaining = output.length - offset;
+      final chunkLength = remaining < bufferLength ? remaining : bufferLength;
+      output.setRange(offset, offset + chunkLength, bytes.copyToDartBytes());
```
jonasfj: lots of allocations here... maybe leave a TODO.

Probably the way out involves understanding:
https://pub.dev/documentation/jni/latest/jni/JByteBuffer-extension-type.html

lib/src/impl_jni/impl_jni.hmac.dart
mfazrinizar: Noted. I kept this PR to small helper patterns for now: arena.copyToJByteArray, copyToDartBytes, and primitive-local helpers like _createMac(arena). I agree broader JCA wrappers may make sense later, but I’d rather wait until AES/RSA/ECDSA show the repeated patterns before introducing that abstraction. Then do the refactoring in a separate PR.

lib/src/impl_jni/impl_jni.utils.dart
mfazrinizar: Changed this to compute chunkLength before calling SecureRandom.nextBytes, so the final partial chunk uses a right-sized JByteArray instead of requesting a full 4096 bytes.

lib/src/impl_jni/impl_jni.utils.dart
mfazrinizar: In this case, avoided copyToDartBytes() here and copy the getRange view directly into the destination buffer, since the destination is already Dart-owned.

I left a TODO for revisiting JByteBuffer-based bulk transfers, but kept it as future work in this PR because SecureRandom.nextBytes only accepts byte[]. JByteBuffer looks more directly useful for MessageDigest/Mac/Cipher paths that have ByteBuffer overloads. Does this look like the right compromise for this PR?

lib/src/impl_jni/impl_jni.hmac.dart
```diff
@@ -14,6 +14,47 @@
 
 part of 'impl_jni.dart';
 
+const _hmacStreamBufferLength = 4096;
```
jonasfj:
```suggestion
const _defaultChunkSize = 4096;
```

or something like that, you'll probably need it for other things where you're streaming byte chunks.

I'd also suggest putting it in `impl_jni.utils.dart`.

None of this important, you can do just do it in a follow up cleanup PR.

lib/src/impl_jni/impl_jni.hmac.dart
```diff
@@ -14,6 +14,47 @@
...
+}
+
+extension _HmacHashMetadata on _HashImpl {
+  String get _hmacJcaName {
+    return switch (_jcaName) {
+      'SHA-1' => 'HmacSHA1',
+      'SHA-256' => 'HmacSHA256',
+      'SHA-384' => 'HmacSHA384',
+      'SHA-512' => 'HmacSHA512',
+      _ => throw AssertionError('Unsupported hash algorithm: $_jcaName'),
```
jonasfj:
```suggestion
      _ => throw AssertionError('Unknown hash algorithm: $_jcaName'),
```

lib/src/impl_jni/impl_jni.utils.dart
```diff
@@ -14,15 +14,108 @@
...
+}
+
+String _jwkEncodeBase64UrlNoPadding(Uint8List data) {
+  final padded = base64Url.encode(data);
+  final paddingStart = padded.indexOf('=');
+  return paddingStart == -1 ? padded : padded.substring(0, paddingStart);
+}
+
+extension _JniArenaByteArray on jni.Arena {
+  jni.JByteArray copyToJByteArray(Iterable<int> data) {
```
jonasfj:
```suggestion
  jni.JByteArray copyToJByteArray(Uint8List data) {
```

Avoid `Iterable<int>` and `List<int>`, prefer `Uint8List` whenever possible.

lib/src/impl_jni/impl_jni.utils.dart
```diff
@@ -14,15 +14,108 @@
...
     return Uint8List.fromList(view);
   }
+
+  void copyRangeToDart(
+    Uint8List destination,
+    int destinationOffset,
+    int length,
+  ) {
+    final bytes = getRange(0, length);
+    final view = bytes.buffer.asUint8List(bytes.offsetInBytes, bytes.length);
```
jonasfj: Try: https://api.dart.dev/dart-typed_data/Uint8List/Uint8List.sublistView.html

lib/src/impl_jni/impl_jni.hmac.dart
jonasfj: 2026

jonasfj (approved):

lib/src/impl_jni/impl_jni.utils.dart
mfazrinizar: Noted. Missed that previously.

HamdaanAliQuatil (approved):

PR #300, opened 2026-06-30
feat(impl_jni): add JCA AES-GCM implementation

mfazrinizar:
## Summary

Adds the JNI/JCA AES-GCM implementation on top of the existing Android JCA backend skeleton.

This PR implements:
- AES raw key import/export helpers for JNI
- AES JWK import/export helpers for JNI
- AES key generation using JCA-backed random bytes
- AES-GCM encryption/decryption through JCA `Cipher("AES/GCM/NoPadding")`
- `GCMParameterSpec` tag length handling
- additional authenticated data
- AES-128-GCM and AES-256-GCM
- focused desktop JNI tests and Android smoke test

## Testing

- `dart run jni:setup`

Verified with:
- `dart analyze`
- `dart test test/impl_jni_aesgcm_test.dart`
- `dart test test/webcrypto_test.dart -p vm -n AES-GCM`
  - Expected current result on desktop SunJCE: fails on the 32/64-bit AES-GCM tag-length vectors with `UnsupportedError: The JCA provider does not support AES-GCM tagLength=...`.
  - This matches the provider limitation documented in this PR. Provider-supported tag lengths pass through the focused JNI coverage, and provider-rejected short tags are translated explicitly.

### DIFF
--- a/lib/src/impl_jni/impl_jni.aesgcm.dart
+++ b/lib/src/impl_jni/impl_jni.aesgcm.dart
@@ -14,18 +14,137 @@
 
 part of 'impl_jni.dart';
 
+const _aesGcmTransformation = 'AES/GCM/NoPadding';
+
+Future<Uint8List> _aesGcmEncryptDecrypt(
+  Uint8List keyData,
+  List<int> data,
+  List<int> iv,
+  List<int>? additionalData,
+  int tagLength,
+  bool isEncrypt,
+) async {
+  final aad = additionalData ?? const <int>[];
+  if (isEncrypt && data.length > (1 << 39) - 256) {
+    // More than this is not allowed by Web Crypto.
+    throw operationError('data may not be more than 2^39 - 256 bytes');
+  }
+  if (tagLength != 32 &&
+      tagLength != 64 &&
+      tagLength != 96 &&
+      tagLength != 104 &&
+      tagLength != 112 &&
+      tagLength != 120 &&
+      tagLength != 128) {
+    throw operationError('tagLength must be 32, 64, 96, 104, 112, 120 or 128');
+  }
+
+  try {
+    return jni.using((arena) {
+      final algorithm = 'AES'.toJString()..releasedBy(arena);
+      final transformation = _aesGcmTransformation.toJString()
+        ..releasedBy(arena);
+      final keyBytes = arena.copyToJByteArray(keyData);
+      final secretKey = SecretKeySpec(keyBytes, algorithm)..releasedBy(arena);
+      final ivBytes = arena.copyToJByteArray(_asUint8List(iv));
+      final parameters = GCMParameterSpec(tagLength, ivBytes)
+        ..releasedBy(arena);
+
+      final cipher = Cipher.getInstance(transformation);
+      if (cipher == null) {
+        throw operationError(
+          'JCA Cipher($_aesGcmTransformation) is unavailable',
+        );
+      }
+      cipher.releasedBy(arena);
+
+      cipher.init$2(
+        isEncrypt ? Cipher.ENCRYPT_MODE : Cipher.DECRYPT_MODE,
+        secretKey,
+        parameters,
+      );
+
+      if (aad.isNotEmpty) {
+        final aadBytes = arena.copyToJByteArray(_asUint8List(aad));
+        cipher.updateAAD(aadBytes);
+      }
+
+      final input = arena.copyToJByteArray(_asUint8List(data));
+      final output = cipher.doFinal$2(input);
+      if (output == null) {
+        throw operationError(
+          'JCA Cipher($_aesGcmTransformation) returned null',
+        );
+      }
+      output.releasedBy(arena);
+      return output.copyToDartBytes();
+    });
+  } on OperationError {
+    rethrow;
+  } on jni.JThrowable catch (e) {
+    final message = e.message;
+    e.release();
+    throw operationError('JCA Cipher($_aesGcmTransformation) failed: $message');
+  }
+}
+
 final class _StaticAesGcmSecretKeyImpl implements StaticAesGcmSecretKeyImpl {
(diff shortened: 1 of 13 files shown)


lib/src/impl_jni/impl_jni.aesgcm.dart
```diff
@@ -14,18 +14,131 @@
...
+    // More than this is not allowed by Web Crypto.
+    throw operationError('data may not be more than 2^39 - 256 bytes');
+  }
+  if (tagLength != 32 &&
+      tagLength != 64 &&
+      tagLength != 96 &&
+      tagLength != 104 &&
+      tagLength != 112 &&
+      tagLength != 120 &&
+      tagLength != 128) {
```
HamdaanAliQuatil, line 37: webcrypto permits all seven tag lengths listed here, but provider support is not uniform. desktop sunjce rejects 32 and 64-bit tags, while the current focused and android tests only exercise 128 bits. 

please add coverage for every accepted tag length on both desktop jni and android, and explain how the backend will preserve the webcrypto contract where the provider rejects 32 or 64.

also, could you share the output from the claimed dart test test/webcrypto_test.dart -p vm -n AES-GCM run? that suite already contains 32 and 64-bit vectors.

lib/src/impl_jni/impl_jni.aesgcm.dart
```diff
@@ -14,18 +14,131 @@
...
+      if (output == null) {
+        throw operationError(
+          'JCA Cipher($_aesGcmTransformation) returned null',
+        );
+      }
+      output.releasedBy(arena);
+      return output.copyToDartBytes();
+    } on OperationError {
+      rethrow;
+    } catch (e) {
```
HamdaanAliQuatil: this catches every dart error, while a caught `jni.JThrowable` retains its global reference until finalization. authentication failure is attacker-triggerable, so this should not rely on eventual gc.

please wrap the complete `jni.using(...)` call, catch `jni.JThrowable` specifically, capture its message, release it in finally, and translate it to `OperationError`. that will also consistently translate exceptions from `SecretKeySpec`, `GCMParameterSpec`, and `Cipher.getInstance`, which currently sit outside this try.

HamdaanAliQuatil (changes requested):

lib/src/impl_jni/impl_jni.aesgcm.dart
jonasfj: If we are not testing all the tagLengths in our existing tests, maybe we should expand those and land that on master branch.

It's certainly possible that for some cases, some platforms have chosen not to support all tagLengths.

it's also possible that we either:
 * Throw `UnsupportedError` on the specific platforms that don't support specific tag legnths, and document this in the documentation string.
 * Implement this by other means, maybe using boringssl, but then rely on treeshaking to avoid dragging those bits along if we don't have to.
 * Implement it in dart... if it's just a few bits we have to move around.

lib/src/impl_jni/impl_jni.aesgcm.dart
mfazrinizar: Addressed.

lib/src/impl_jni/impl_jni.aesgcm.dart
mfazrinizar: I confirmed the provider issue locally. SunJCE rejects 32/64 with `Unsupported TLen value`. I also found the shared AES-GCM test runner currently reads `length` while the fixtures use `tagLength`, so I’ll fix that in a separate PR to master.

mfazrinizar: Rebased after #312 and #313.

test/impl_jni_aesgcm_test.dart
```diff
@@ -0,0 +1,248 @@
...
+      for (final tagLength in providerDependentTagLengths) {
+        final ffiCiphertext = await ffiKey.encryptBytes(
+          plaintext,
+          iv,
+          additionalData: additionalData,
+          tagLength: tagLength,
+        );
+
+        try {
+          final ciphertext = await key.encryptBytes(
```
HamdaanAliQuatil, line 195: this test covers the short-tag encrypt path, but not the failing path from the shared suite: decrypting an existing 32/64-bit ciphertext.

if `encryptBytes` throws `UnsupportedError`, this test never calls `decryptBytes(ffiCiphertext, ...)`. users can still receive short-tag ciphertext generated elsewhere, so please exercise both directions independently for provider dependent tag lengths. each direction can either match ffi or throw `UnsupportedError` with the tag length in the message.

HamdaanAliQuatil (changes requested):
i reran this after `dart run jni:setup`.

 `dart analyze` passes, and `dart test test/impl_jni_aesgcm_test.dart` passes.

`dart test test/webcrypto_test.dart -p vm -n AES-GCM` does not pass on desktop sunjce. it fails on the 32/64-bit shared vectors with `UnsupportedError: The JCA provider does not support AES-GCM tagLength=...`. that matches the provider limitation we discussed, but the pr body currently lists this command as verified/ passing.

please update the testing section to be explicit: focused jni aes-gcm tests pass, while the shared aes-gcm suite currently fails on desktop providers that reject 32/64-bit tags. also add the decrypt-side short-tag coverage in the focused test.

test/impl_jni_aesgcm_test.dart
mfazrinizar: Updated the PR body testing section to state that explicitly instead of listing it as passing.

Also, the provider-dependent short-tag test now exercises encrypt and decrypt independently. The decrypt side uses FFI-generated 32/64-bit ciphertext, so JCA providers either decrypt it successfully or throw `UnsupportedError` with the tag length in the message.

HamdaanAliQuatil (approved):

PR #303, opened 2026-07-02
Fix HMAC JWK import length handling

### LINKED ISSUE #302: HmacSecretKey.importJsonWebKey ignores length
### Description

`HmacSecretKey.importJsonWebKey` accepts a `length` parameter, but the public wrapper does not forward it to the backend. As a result, importing the same HMAC key material as raw bytes and as JWK can produce different key bytes and different signatures whenever `length` trims bits from the final byte.

This breaks the package's cross-format parity: `HmacSecretKey.importRawKey(..., length: n)` honors the requested HMAC key length, while `HmacSecretKey.importJsonWebKey(..., length: n)` silently uses the full JWK `k` value.

### Expected behavior

JWK import should honor `length` the same way raw import does. In the example above, both imports should export `[255, 128]`, and signatures produced by the two keys should match.

### Actual behavior

The `length` argument is dropped in `HmacSecretKey.importJsonWebKey`, so the JWK-backed key keeps the untrimmed final byte and signs/verifies with different key material.

### Notes

A fix should also validate JWK import length consistently with raw imports so the native backend does not silently accept invalid lengths.

harrshita123:
Fixes #302.

### What changed
- Pass `length` through HMAC JWK import.
- Apply `length` after decoding `jwk.k`, matching Web Crypto HMAC import steps.
- Keep raw and JWK import behavior consistent for non-byte-aligned lengths.
- Add regression coverage for raw/JWK parity and invalid HMAC key length.

### Testing
- `/Users/apple/flutter/bin/dart run lib/src/testing/regression/issue_302_hmac_jwk_length.dart`
- `/Users/apple/flutter/bin/dart test test/webcrypto_test.dart`
- `/Users/apple/flutter/bin/dart analyze`

### DIFF
--- a/lib/src/testing/regression/issue_302_hmac_jwk_length.dart
+++ b/lib/src/testing/regression/issue_302_hmac_jwk_length.dart
@@ -22,10 +22,15 @@ List<({String name, Future<void> Function() test})> tests() {
   void test(String name, Future<void> Function() fn) =>
       tests.add((name: name, test: fn));
 
-  test('Hmac: importJsonWebKey accepts matching length', () async {
-    final keyData = [0xff, 0x80];
-    final jwk = {'kty': 'oct', 'alg': 'HS256', 'k': '_4A'};
+  test('Hmac: importJsonWebKey applies length like importRawKey', () async {
+    final keyData = [0xff, 0xe0];
+    final jwk = {'kty': 'oct', 'alg': 'HS256', 'k': '_-A'};
 
+    final rawKey = await HmacSecretKey.importRawKey(
+      keyData,
+      Hash.sha256,
+      length: 9,
+    );
     final jwkKey = await HmacSecretKey.importJsonWebKey(
       jwk,
       Hash.sha256,
@@ -33,21 +38,15 @@ List<({String name, Future<void> Function() test})> tests() {
     );
 
     check(
-      equalBytes(await jwkKey.exportRawKey(), keyData),
-      'JWK import should preserve key data that matches length',
+      equalBytes(await rawKey.exportRawKey(), await jwkKey.exportRawKey()),
+      'JWK import should zero unused bits the same way as raw import',
     );
-  });
 
-  test('Hmac: importJsonWebKey rejects non-zero unused bits', () async {
-    final jwk = {'kty': 'oct', 'alg': 'HS256', 'k': '_-A'};
-
-    bool threw = false;
-    try {
-      await HmacSecretKey.importJsonWebKey(jwk, Hash.sha256, length: 9);
-    } on FormatException {
-      threw = true;
-    }
-    check(threw, 'Should throw FormatException for non-zero unused bits');
+    final data = [1, 2, 3, 4];
+    check(
+      equalBytes(await rawKey.signBytes(data), await jwkKey.signBytes(data)),
+      'JWK import should use the requested HMAC key length for signing',
+    );
   });
 
   test('Hmac: importJsonWebKey validates length', () async {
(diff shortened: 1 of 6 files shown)


lib/src/webcrypto/webcrypto.hmac.dart
```diff
@@ -162,26 +174,19 @@ final class HmacSecretKey {
...
     // These limitations are given in Web Cryptography Spec:
     // https://www.w3.org/TR/WebCryptoAPI/#hmac-operations
-    if (length != null && length > keyData.length * 8) {
-      throw ArgumentError.value(
-          length, 'length', 'must be less than number of bits in keyData');
-    }
-    if (length != null && length <= (keyData.length - 1) * 8) {
-      throw ArgumentError.value(
+    if (length != null && jwk['k'] is String) {
+      _checkLength(
```
jonasfj: >   'must be greater than number of bits in keyData - 8, you can attain '
>   'the same effect by removing bytes from keyData',

This is actually not an argument error. and the issue in not in `keyData`.

It's `jwk` having a format exception potentially, it certain doesn't live up to the expected length.

Ideally, this should live in the implementation instead of in there.

lib/src/testing/regression/issue_302_hmac_jwk_length.dart
```diff
@@ -0,0 +1,65 @@
...
+    );
+    final jwkKey = await HmacSecretKey.importJsonWebKey(
+      jwk,
+      Hash.sha256,
+      length: 9,
+    );
+
+    check(
+      equalBytes(await rawKey.exportRawKey(), await jwkKey.exportRawKey()),
+      'JWK import should zero unused bits the same way as raw import',
```
jonasfj, line 42: JWK import should never modify the key, only validate it.

lib/src/webcrypto/webcrypto.hmac.dart
harrshita123: I updated this so the JWK length validation now lives in the backend implementations instead of the public wrapper.

The validation now treats this as a JWK format/content issue and throws `FormatException` for `jwk.k` when it does not match the expected length, rather than using `ArgumentError` for `keyData`.

lib/src/testing/regression/issue_302_hmac_jwk_length.dart
harrshita123: I updated the implementation and regression tests so JWK import no longer mutates the key material.

Now `length` is treated only as an expected key length: the decoded `jwk.k` must already match that length, including zero unused trailing bits for non-byte-aligned lengths. If it does not match, the JWK is rejected instead of being modified.

harrshita123:
Sorry, I accidentally clicked the close button. I have reopened the PR.

harrshita123:
@jonasfj I resolved all conflicts.

HamdaanAliQuatil (changes requested):
I checked this against the normative HMAC import steps and left the details in #302. Unlike hash or curve, length is applied after the JWK has been decoded: the imported key consists of the first length bits.

That means the current validation-only implementation, particularly the test rejecting non-zero discarded bits, would encode behavior that differs from Web Crypto. I think we should settle the expected behavior in #302 before merging this.

harrshita123:
Thanks for the clarification and for checking the HMAC import steps against the Web Crypto spec.

I updated the PR so JWK import now follows the same behavior as raw import: it decodes `jwk.k` first, then applies `length` by using the first `length` bits. I also removed the validation-only behavior that rejected non-zero discarded bits, since that would not match the spec.

The regression test now covers the raw/JWK parity case for `{ k: "_-A" }` with `length: 9`, and the focused regression test, `dart analyze`, and `test/webcrypto_test.dart` are passing.

HamdaanAliQuatil (approved):

PR #304, opened 2026-07-02
feat(impl_jni): add JCA AES-CBC implementation

mfazrinizar:
## Summary

Adds the JNI/JCA AES-CBC implementation on top of the existing Android JCA backend skeleton.

This PR implements:
- AES-CBC raw key import/export
- AES-CBC JWK import/export
- AES-CBC key generation using the shared AES helper
- byte and stream encryption/decryption
- JCA `Cipher("AES/CBC/PKCS5Padding")` wiring with `IvParameterSpec`
- focused desktop JNI tests
- Android integration smoke test through the public API

`PKCS5Padding` is the JCA transformation name used for AES-CBC PKCS#7-style block padding.

## Testing

- `dart run jni:setup`

Verified with:
- `dart analyze`
- `dart test test/impl_jni_aescbc_test.dart`
- `dart test test/webcrypto_test.dart -p vm -n AES-CBC`
- `flutter test integration_test/jni_aescbc_test.dart -d emulator-name`

### DIFF
--- a/lib/src/impl_jni/impl_jni.aescbc.dart
+++ b/lib/src/impl_jni/impl_jni.aescbc.dart
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


HamdaanAliQuatil (approved):

lib/src/impl_jni/impl_jni.aescbc.dart
```diff
@@ -14,18 +14,171 @@
...
+      final input = arena.copyToJByteArray(bytes);
+      final output = jni.JByteArray(cipher!.getOutputSize(bytes.length))
+        ..releasedBy(arena);
+      final outputLength = cipher.doFinal$4(input, 0, bytes.length, output);
+      return output.copyToDartBytes(length: outputLength);
+    });
+  } on OperationError {
+    rethrow;
+  } on jni.JThrowable catch (e) {
+    final message = e.message;
```
HamdaanAliQuatil: same throwable lifetime issue as https://github.com/google/webcrypto.dart/pull/300: invalid padding/decryption failures are attacker-triggerable, so we should not rely on finalizer cleanup if reading e.message throws.

please copy the message inside a try, release the JThrowable in finally, and apply the same pattern to the stream catch block below.

HamdaanAliQuatil (changes requested):

lib/src/impl_jni/impl_jni.aescbc.dart
mfazrinizar: Fixed. Both AES-CBC byte and stream JThrowable paths now use a helper that copies the message inside try/finally and releases the throwable in finally, matching the AES-GCM pattern from #300.

HamdaanAliQuatil (approved):

PR #309, opened 2026-07-05
Fix AES-CTR large chunk handling

### LINKED ISSUE #307: bug: native AES-CTR corrupts stream chunks larger than 4096 bytes
### Description

The native FFI AES-CTR implementation silently corrupts data when a source chunk is larger than 4096 bytes and is not an exact multiple of 4096.

For a 4097-byte plaintext, `encryptBytes()` returns 8192 bytes. Decrypting it does not recover the original plaintext. With assertions enabled, the operation instead fails at `assert(i == M)`.

### Reproduction

```dart

Future<void> main() async {
  final key = await AesCtrSecretKey.importRawKey(List<int>.filled(16, 0));
  final plaintext = List<int>.generate(4097, (i) => i & 0xff);
  final counter = List<int>.filled(16, 0);

  final ciphertext = await key.encryptBytes(plaintext, counter, 128);
  final decrypted = await key.decryptBytes(ciphertext, counter, 128);

  print('plaintext length:  ${plaintext.length}');
  print('ciphertext length: ${ciphertext.length}');
  print('decrypted length:  ${decrypted.length}');
}
```

Output:

```text
plaintext length:  4097
ciphertext length: 8192
decrypted length:  8192
```

With `dart --enable-asserts run`:

```text
Failed assertion: 'i == M': is not true.
```

### Expected behavior

AES-CTR ciphertext must have the same length as the plaintext, and decrypting it with the same key and counter must recover the original bytes.

### Root cause

In `lib/src/impl_ffi/impl_ffi.aesctr.dart`, the inner buffering loop uses:

```dart
final N = math.min(M, bufSize);
```

After the first 4096-byte block, this still passes 4096 to `EVP_CipherUpdate`, even when fewer bytes remain. The unused buffer tail
contains stale data.

The chunk size should account for bytes already consumed:

```dart
final N = math.min(M - i, bufSize);
```

### Impact

- Affects the native FFI backend.
- Affects `encryptBytes()` and `decryptBytes()`.
- Also affects streaming operations whenever an individual source chunk or
  counter-wrap segment exceeds 4096 bytes with a partial final block.
- Causes silent cryptographic data corruption when assertions are disabled.

### Suggested regression coverage

Test a 4097-byte input and verify:

1. Ciphertext length equals plaintext length.
2. Encryption/decryption round-trips successfully.
3. Both byte and streaming APIs behave identically.

### Environment

- `webcrypto` 0.6.1
- Current `master` contains the affected loop
- Dart 3.12.1
- macOS arm64

The existing test suite passes because its generated AES-CTR plaintext is limited to small inputs.

harrshita123:
Fixes #307.

### Summary

- account for bytes already consumed when selecting the next AES-CTR input chunk
- add regression coverage for a 4097-byte input
- verify byte encryption/decryption round-trips and matches streaming encryption

### Testing

- `dart format --output none --set-exit-if-changed lib/src/impl_ffi/impl_ffi.aesctr.dart test/aes_ctr_counter_wrap_test.dart`
- `dart test -p vm test/aes_ctr_counter_wrap_test.dart`
- `dart test -p chrome test/aes_ctr_counter_wrap_test.dart`
- `dart test -p vm`

### DIFF
--- a/test/aes_ctr_counter_wrap_test.dart
+++ b/test/aes_ctr_counter_wrap_test.dart
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import 'dart:async';
 import 'dart:typed_data';
 
 import 'package:test/test.dart';
@@ -36,5 +37,22 @@ void main() {
     final ciphertextB = await key.encryptBytes(Uint8List(16), counterB, 32);
 
     expect(ciphertextA.sublist(16, 32), isNot(equals(ciphertextB)));
+  });
+
+  test('AES-CTR handles input larger than the internal buffer', () async {
+    final key = await AesCtrSecretKey.importRawKey(Uint8List(16));
+    final plaintext = Uint8List(4097);
+    final counter = Uint8List(16);
+
+    final ciphertext = await key.encryptBytes(plaintext, counter, 128);
+    final decrypted = await key.decryptBytes(ciphertext, counter, 128);
+    final streamedCiphertext = await key
+        .encryptStream(Stream.value(plaintext), counter, 128)
+        .expand((chunk) => chunk)
+        .toList();
+
+    expect(ciphertext, hasLength(plaintext.length));
+    expect(decrypted, equals(plaintext));
+    expect(streamedCiphertext, equals(ciphertext));
   });
 }
--- a/test/aes_ctr_counter_wrap_test.dart
+++ b/test/aes_ctr_counter_wrap_test.dart
@@ -39,20 +39,36 @@ void main() {
     expect(ciphertextA.sublist(16, 32), isNot(equals(ciphertextB)));
   });
 
-  test('AES-CTR handles input larger than the internal buffer', () async {
+  test('AES-CTR keeps large byte encryption length stable', () async {
     final key = await AesCtrSecretKey.importRawKey(Uint8List(16));
     final plaintext = Uint8List(4097);
     final counter = Uint8List(16);
 
     final ciphertext = await key.encryptBytes(plaintext, counter, 128);
     final decrypted = await key.decryptBytes(ciphertext, counter, 128);
-    final streamedCiphertext = await key
-        .encryptStream(Stream.value(plaintext), counter, 128)
-        .expand((chunk) => chunk)
-        .toList();
 
     expect(ciphertext, hasLength(plaintext.length));
     expect(decrypted, equals(plaintext));
+  });
+
+  test('AES-CTR stream encryption matches large byte encryption', () async {
+    final key = await AesCtrSecretKey.importRawKey(Uint8List(16));
+    final plaintext = Uint8List(4097);
+    final counter = Uint8List(16);
+
+    final ciphertext = await key.encryptBytes(plaintext, counter, 128);
+    final streamedCiphertext = await _collectBytes(
+      key.encryptStream(Stream.value(plaintext), counter, 128),
+    );
+
     expect(streamedCiphertext, equals(ciphertext));
   });
+}
+
+Future<Uint8List> _collectBytes(Stream<List<int>> stream) async {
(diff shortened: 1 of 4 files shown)


test/aes_ctr_counter_wrap_test.dart
```diff
@@ -37,4 +38,21 @@ void main() {
...
+    final counter = Uint8List(16);
+
+    final ciphertext = await key.encryptBytes(plaintext, counter, 128);
+    final decrypted = await key.decryptBytes(ciphertext, counter, 128);
+    final streamedCiphertext = await key
+        .encryptStream(Stream.value(plaintext), counter, 128)
+        .expand((chunk) => chunk)
+        .toList();
+
+    expect(ciphertext, hasLength(plaintext.length));
```
jonasfj, line 50: Is this to be expected?

test/aes_ctr_counter_wrap_test.dart
```diff
@@ -37,4 +38,21 @@ void main() {
...
+    final key = await AesCtrSecretKey.importRawKey(Uint8List(16));
+    final plaintext = Uint8List(4097);
+    final counter = Uint8List(16);
+
+    final ciphertext = await key.encryptBytes(plaintext, counter, 128);
+    final decrypted = await key.decryptBytes(ciphertext, counter, 128);
+    final streamedCiphertext = await key
+        .encryptStream(Stream.value(plaintext), counter, 128)
+        .expand((chunk) => chunk)
+        .toList();
```
jonasfj: I'd suggest making a separate test case.

Also use `BytesBuilder` or something other than `.expand .toList()` that's a bit wild as a way to work with bytes 🤣

test/aes_ctr_counter_wrap_test.dart
```diff
@@ -37,4 +38,21 @@ void main() {
 
     expect(ciphertextA.sublist(16, 32), isNot(equals(ciphertextB)));
   });
+
+  test('AES-CTR handles input larger than the internal buffer', () async {
```
jonasfj: Try running this:
https://github.com/google/webcrypto.dart/blob/master/lib/src/testing/webcrypto/aesctr.dart#L66

with `maxPlaintext: 4097`, then you should be able to a JSON test data case we can just embed in https://github.com/google/webcrypto.dart/blob/master/lib/src/testing/webcrypto/aesctr.dart#L75

I'll grant that maybe `4097` is a bit large and please do use multi-line strings or something like that write it in the code so that it's elegant :D

But might be nice to have covered on all platforms, granted this test case is also fine, I'm guessing this runs everything.

jonasfj (approved):

jonasfj:
Oh, please do a CHANGELOG update too

test/aes_ctr_counter_wrap_test.dart
harrshita123: Yes, this is expected for AES-CTR. It does not add padding, so the ciphertext length should match the plaintext length.

test/aes_ctr_counter_wrap_test.dart
harrshita123: I split this into separate test cases and replaced the stream flattening with BytesBuilder.

test/aes_ctr_counter_wrap_test.dart
harrshita123: I added a 4097-byte AES-CTR test vector to the shared test data so this path gets broader coverage.

harrshita123:
> Oh, please do a CHANGELOG update too

I added a CHANGELOG entry for the native AES-CTR large-input fix.

HamdaanAliQuatil (approved):
This is a good catch. AES-CTR should preserve input length, so the native path producing extra bytes here is real data corruption.

Reviewed the fix and regression coverage. I also validated the focused AES-CTR test and the VM webcrypto suite locally. Thanks for the fix 🚀

Issue #54, opened 2022-11-02
Consider adopting test vectors from chromium


jonasfj:
There are some JSON test vectors in chromium, it might be nice to embed these JSON files in `third_party/chromium/...` and create a test that embeds the JSON test vectors, to see that we get the same results across platforms.

https://source.chromium.org/chromium/chromium/src/+/main:components/test/data/webcrypto/bad_ec_keys.json

harrshita123:
Hi @jonasfj, I’d like to work on #54. I propose a bounded first PR around Chromium’s bad_ec_keys.json: vendor it at a pinned Chromium revision with source/license metadata, add data-driven negative tests for applicable ECDH/ECDSA JWK, SPKI, and PKCS8 imports, and run them on both FFI and browser backends. I’ll assert that imports are rejected rather than matching Chromium’s exact error strings.  
I noticed #135 also adds import-exception support to TestRunner. Would you prefer this work to reuse/supersede that approach, or should I create a focused standalone regression runner?

jonasfj:
a tool/ script to vendor the required files sounds decent.

Probably it's preferable to load the files from disk at runtime, so they live in the third_party/chromium/ folder.

But that also means that browser tests are hard to do. Maybe spawn hybrid make it work. Otherwise, I'd argue the most important target is ffi backend.

harrshita123:
> a tool/ script to vendor the required files sounds decent.
> 
> Probably it's preferable to load the files from disk at runtime, so they live in the third_party/chromium/ folder.
> 
> But that also means that browser tests are hard to do. Maybe spawn hybrid make it work. Otherwise, I'd argue the most important target is ffi backend.

Thanks. I will make the first PR FFI-focused: add a pinned Chromium revision and vendoring script for bad_ec_keys.json, store the vector and attribution under third_party/chromium/, and add VM/FFI tests that load it from disk. I will leave browser execution with spawnHybridUri for a follow-up after the native tests work.

PR #314, opened 2026-07-11
test(ffi): adopt Chromium invalid EC key vectors

harrshita123:
Addresses part of #54.

## Summary

- Vendor Chromium's `bad_ec_keys.json` corpus and license at a pinned revision.
- Add a repeatable script for updating the vendored files.
- Add VM-only regression tests covering all 21 malformed EC keys across JWK, SPKI, and PKCS8 imports.

The test reads the vectors from `third_party/chromium/` at runtime and keeps the upstream file unchanged, including its comment-only lines. It checks that native ECDSA imports reject each malformed key with a `FormatException` without depending on Chromium's exact error text.

Browser execution is intentionally left for a follow-up, as discussed in #54.

## Testing

- `dart format --output none --set-exit-if-changed .`
- `dart test test/ffi/chromium_bad_ec_keys_test.dart`
- `dart test --exclude-tags browser-interop -p vm,chrome -c dart2js,dart2wasm`


mosuem:
I made some suggestions in the `test/chromium-bad-ec-vectors` branch - WDYT @harrshita123 ? Just a few nits to make it more maintainable

test/ffi/chromium_bad_ec_keys_test.dart
```diff
@@ -0,0 +1,134 @@
...
+          throwsA(isA<FormatException>()),
+          reason: 'Chromium expectation: ${vector.expectedError}',
+        );
+      });
+    }
+  });
+}
+
+List<_BadEcKeyVector> _loadTestVectors() {
+  final file = File(_testVectorPath);
```
jonasfj: Use Isolate.resolvePackageUriSync(package:webcrypto) to find where this is..

harrshita123: I updated this to resolve `package:webcrypto/webcrypto.dart` with `Isolate.resolvePackageUriSync` and derive the vector path from that URI. This also removes the package-root working-directory requirement. The focused test still passes all 21 vectors.

harrshita123:
I updated the test to locate the vendored Chromium vectors from `package:webcrypto` with `Isolate.resolvePackageUriSync`, so it no longer depends on the current working directory. The focused test passes all 21 vectors, and `dart analyze` reports no issues.

PR #315, opened 2026-07-12
feat: publish prebuilt native assets

HamdaanAliQuatil:
This builds trusted desktop native libraries during the pub.dev release workflow and includes them in the published package without committing binaries to the repository.

What changes:

- `hook/build.dart` uses a matching packaged library from `prebuilt/<target>/`, falling back to the existing CMake source build when no prebuilt is available.
- Consumers and CI can set `build_from_source: true` to bypass an available prebuilt explicitly.
- The publish workflow builds Linux x64, macOS arm64/x64, and Windows x64 artifacts, validates them, assembles the package, and publishes only from a version tag.
- Linux is built on Ubuntu 20.04 and checked for a maximum GLIBC 2.31 requirement. macOS deployment targets are checked at 10.15 for x64 and 11.0 for arm64.
- Pull requests run the same four-platform build and package validation. The validation exercises both the prebuilt and forced source-build paths and reports the exact compressed pub archive size.

Local validation:

- `dart analyze hook/build.dart`
- `dart format --output=none --set-exit-if-changed hook/build.dart`
- `dart test -p vm test/aes_ctr_counter_wrap_test.dart`

CI assembled the four-platform pub archive at 6.34 MiB compressed (6,646,685 bytes). The four native libraries are 8.61 MiB uncompressed (9,032,224 bytes).


hook/build.dart
```diff
@@ -69,6 +89,31 @@ Future<void> main(List<String> args) async {
...
+
+String _targetName(BuildInput input) {
+  final code = input.config.code;
+  final os = code.targetOS;
+  final arch = code.targetArchitecture;
+  if (os == OS.iOS) {
+    return '${os.name}-${code.iOS.targetSdk.type}-${arch.name}';
+  }
+  return '${os.name}-${arch.name}';
+}
```
jonasfj: This is parsed from YAML and represented as JSON, I don't think `'true'` is something we need to compare to.

Also you could use extension methods to keep it pretty.
```suggestion
extension on BuildInput {
  File get prebuiltAsset {
    final libraryFileName = config.code.targetOS.dylibFileName(_libraryName);
    return File.fromUri(
      input.packageRoot.resolve(
        'prebuilt/$targetName/$libraryFileName',
      ),
    );
  }
  
  String get targetName {
    final code = config.code;
    final os = code.targetOS;
    final arch = code.targetArchitecture;
    if (os == OS.iOS) {
      return '${os.name}-${code.iOS.targetSdk.type}-${arch.name}';
    }
    return '${os.name}-${arch.name}';
  }
}

extension on HookInputUserDefines {
  bool get forceSourceBuild => this[_forceSourceBuildDefine] == true;
}
```

hook/build.dart
```diff
@@ -19,6 +19,8 @@ import 'package:hooks/hooks.dart';
 import 'package:native_toolchain_cmake/native_toolchain_cmake.dart';
 
 const _assetName = 'webcrypto.dart';
+const _libraryName = 'webcrypto';
+const _forceSourceBuildDefine = 'force_source_build';
```
jonasfj: maybe just:

```suggestion
const _forceSourceBuildDefine = 'build_from_source';
```

prebuilt/README.md
```diff
@@ -0,0 +1,59 @@
+# Prebuilt Native Assets
```
jonasfj: I don't think we should commit this directory to git.

I do think it makes sense for us to document, either to future selves or others how binary artifacts are included in the package.

So I'd suggest putting this file in `doc/prebuilt-native-assets.md`

prebuilt/README.md
```diff
@@ -0,0 +1,59 @@
...
+  user_defines:
+    webcrypto:
+      force_source_build: true
+```
+
+## Policy
+
+Do not accept arbitrary binary blobs from contributor PRs. Prebuilt libraries
+should be generated by trusted project automation or another reviewable release
+process.
```
jonasfj: I don't think we need to state this.

Instead I think we should state that prebuilt binaries are build in github actions as part of the flow that publishes to pub.dev

.github/workflows/build-prebuilt-assets.yml
```diff
@@ -0,0 +1,58 @@
+name: build-prebuilt-assets
+
+on:
+  workflow_dispatch:
```
jonasfj: We are thinking that the long term fix, is to put this into:
https://github.com/google/webcrypto.dart/blob/master/.github/workflows/publish.yml

mosuem:
Hi @HamdaanAliQuatil , would you mind taking a look at #325 ? I implemented basically the same thing, based on https://github.com/unicode-org/icu4x/blob/main/ffi/dart/hook/build.dart, which works quite well.

PR #316, opened 2026-07-12
feat(impl_jni): add JCA AES-CTR implementation

mfazrinizar:
## Summary

Adds the JNI/JCA AES-CTR implementation on top of the existing Android JCA backend skeleton.

This PR implements:
- AES-CTR raw key import/export
- AES-CTR JWK import/export
- AES-CTR key generation using the shared AES helper
- byte and stream encryption/decryption
- JCA `Cipher("AES/CTR/NoPadding")` wiring with `IvParameterSpec`
- WebCrypto counter-length rollover handling so counter increments do not carry into nonce bits
- focused desktop JNI tests
- Android integration smoke test through the public API

## Testing

- `dart run jni:setup`

Verified with:
- `dart analyze`
- `dart test test/impl_jni_aesctr_test.dart`
- `dart test test/webcrypto_test.dart -p vm -n AES-CTR`
- `dart test test/aes_ctr_counter_wrap_test.dart`
- `flutter test integration_test/jni_aesctr_test.dart -d emulator-name`

### DIFF
--- a/lib/src/impl_jni/impl_jni.aesctr.dart
+++ b/lib/src/impl_jni/impl_jni.aesctr.dart
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


lib/src/impl_jni/impl_jni.aesctr.dart
```diff
@@ -14,18 +14,264 @@
...
+  final remainder = length % 8;
+  if (remainder != 0) {
+    final index = wrapped.length - counterBytes - 1;
+    wrapped[index] &= 0xff & (0xff << remainder);
+  }
+  return wrapped;
+}
+
+int _minBigIntAndInt(BigInt value, int limit) {
+  // Avoid converting huge counter-space values, such as 2^128,
```
HamdaanAliQuatil: nit: there is trailing whitespace here, which causes dart format --output=none --set-exit-if-changed and git diff --check to fail. please remove it.

HamdaanAliQuatil (changes requested):

lib/src/impl_jni/impl_jni.aesctr.dart
mfazrinizar: Missed checking dart format, thanks. Removed it.

HamdaanAliQuatil (approved):

PR #317, opened 2026-07-13
chore(jca): add RSA JNI bindings

mfazrinizar:
## Summary

Adds JNIgen coverage for the JCA RSA APIs needed by the upcoming RSASSA-PKCS1-v1_5 implementation and reusable RSA key handling.

The first commit updates `jnigen.yaml`. The second commit contains only the regenerated bindings. There is no JCA JNI backend implementation is included.

## Added bindings

| Binding | Purpose |
| --- | --- |
| `java.math.BigInteger` | Represents RSA modulus, exponent, and CRT integer parameters required by JCA key specifications. |
| `java.security.Key` | Provides the common encoded-key interface and correctly types existing `Cipher` and `Mac` key parameters. |
| `java.security.KeyFactory` | Imports PKCS#8, SPKI, and RSA key-spec values into JCA key objects. |
| `java.security.KeyPair` | Exposes generated RSA private and public keys. |
| `java.security.KeyPairGenerator` | Generates RSA key pairs from modulus length and public exponent parameters. |
| `java.security.Signature` | Provides RSASSA signing, verification, and streaming update operations. |
| `java.security.interfaces.RSAKey` | Provides the shared RSA modulus interface. |
| `java.security.interfaces.RSAPublicKey` | Exposes modulus and public exponent for public JWK export. |
| `java.security.interfaces.RSAPrivateKey` | Exposes the private exponent inherited by RSA CRT keys. |
| `java.security.interfaces.RSAPrivateCrtKey` | Exposes private exponent, primes, CRT exponents, and coefficient for private JWK export. |
| `java.security.spec.PKCS8EncodedKeySpec` | Imports DER-encoded PKCS#8 private keys. |
| `java.security.spec.X509EncodedKeySpec` | Imports DER-encoded SPKI public keys. |
| `java.security.spec.RSAKeyGenParameterSpec` | Configures RSA modulus length and public exponent during generation. |
| `java.security.spec.RSAPublicKeySpec` | Constructs public RSA keys from JWK `n` and `e` values. |
| `java.security.spec.RSAPrivateCrtKeySpec` | Constructs private RSA keys from JWK `n`, `e`, `d`, `p`, `q`, `dp`, `dq`, and `qi` values. |
| `javax.crypto.SecretKey` | Preserves static compatibility for existing AES and HMAC bindings after `Cipher` and `Mac` parameters become typed as `java.security.Key`. |

RSA-PSS and RSA-OAEP parameter-spec bindings are intentionally excluded. They can be added with their respective primitive PRs once the required provider behavior is known.

Adding the Java key hierarchy also refines some existing generated `Cipher`, `Mac`, and `SecretKeySpec` signatures from generic `JObject` values to their actual `Key`/`SecretKey` interfaces. This accounts for the rewritten sections in the generated file.

## Testing

For regenration, run from the package root:
- `bash tool/update-jca-bindings.sh`

Desktop JNI setup, if not already available:
- `dart run jni:setup`

Verified with:
- `dart analyze`
- `dart test test/impl_jni_digest_test.dart`
- `dart test test/impl_jni_hmac_test.dart`
- `dart test test/impl_jni_aesgcm_test.dart`
- `dart test test/impl_jni_aescbc_test.dart`

### DIFF
--- a/lib/src/third_party/jca/README.md
+++ b/lib/src/third_party/jca/README.md
@@ -12,12 +12,14 @@ bash tool/update-jca-bindings.sh
 Requirements:
 
 - Flutter SDK
+- Java Development Kit (JDK) 17 (`java -version` must resolve to JDK 17)
 - Android SDK Platform 36 installed and discoverable by Android tooling
   (`ANDROID_SDK_ROOT` can be set explicitly if needed)
 
-The JNIgen config is pinned to Android API 36 so the checked-in generated
-bindings are reproducible across machines. Do not change the API level locally
-to match an installed SDK; install Android SDK Platform 36 instead.
+The known reproducible generation environment uses JDK 17. The JNIgen config
+is pinned to Android API 36 so the checked-in generated bindings are
+reproducible across machines. Use JDK 17 and Android SDK Platform 36 for
+regeneration rather than substituting newer locally installed versions.
 
 The generated output is:
 
--- a/lib/src/third_party/jca/generated_bindings.dart
+++ b/lib/src/third_party/jca/generated_bindings.dart
@@ -40,6 +40,2987 @@ import 'package:jni/jni.dart' as jni$_;
 
 const _$jniVersionCheck = jni$_.JniVersionCheck(1, 0);
 
+/// from: `java.math.BigInteger`
+extension type BigInteger._(jni$_.JObject _$this)
+    implements jni$_.JNumber, jni$_.JObject {
+  static final _class = jni$_.JClass.forName(r'java/math/BigInteger');
+
+  /// The type which includes information such as the signature of this class.
+  static const jni$_.JType<BigInteger> type = $BigInteger$Type$();
+  static final _id_ZERO = _class.staticFieldId(
+    r'ZERO',
+    r'Ljava/math/BigInteger;',
+  );
+
+  /// from: `static public final java.math.BigInteger ZERO`
+  /// The returned object must be released after use, by calling the [release] method.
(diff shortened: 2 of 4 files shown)


HamdaanAliQuatil (changes requested):
i’m blocking on regeneration reproducibility for now.

i ran the documented command with jnigen 0.16.0 and android api 36. jnigen currently passes the discovered android.jar as a directory uri, which produces an android.jar\ classpath on windows and fails. after locally correcting only that upstream path handling to use a file uri, generation succeeds against api 36, but it does not reproduce this file: roughly 2,400 lines change, mostly member ordering and overload numbering, with at least one surface difference (cipher.tostring()).

could you share the os/jdk version used for generation and the complete summarizer log line beginning with execute java -jar ... -c? please also run generation a second time and confirm that git status --short is clean.

since this pr consists almost entirely of generated code, we need to establish that android.jar is actually being consumed rather than methods being resolved from the host jdk.

HamdaanAliQuatil:
non-blocking nit: please document jdk 17 alongside the android sdk requirement so the known reproducible environment is explicit.

mfazrinizar:
> non-blocking nit: please document jdk 17 alongside the android sdk requirement so the known reproducible environment is explicit.

Updated. I documented JDK 17 as part of the reproducible generation environment in both the JCA bindings README and generator script, including that java -version should resolve to JDK 17. Is that enough?

HamdaanAliQuatil (approved):

Issue #283, opened 2026-06-10
Move dartdoc examples to separate files


jonasfj:
As of recent `dartdoc` versions it's now possible to do:
```dart
/// My documentation string...
///
/// {@example /path/to/example/file.dart}
void someMethod() { ... }
```

Doing this makes it easy to run `dart analyze` on examples and use `dart format` on examples.

I propose that we refactor the current repository layout such that we have:
```dart
example/
  flutter_webcrypto_demo/
    ... everything that's currently in example/
  webcrypto/
    hmac/
      import_raw_key.dart
      ...
    <class>/
      <identifier>.dart
```

And then we use `/// {@example /example/webcrypto/hmac/import_raw_key.dart}` in our API docs.

This should not land as one large PR, please do:
 * [x] One PR moving `example/` to `example/flutter_webcrypto_demo/`
    * One commit that moves files
    * One commit that fixes CI
    * Any further cleanup of names, do that in an other PR
 * [ ] A PR for each primitive, so we don't have too big PRs to review

Samarborkar:
Hi @jonasfj , I'd like to work on this issue. I'll start with the first PR — moving example/ to example/flutter_webcrypto_demo/. Let me know if there's anything specific to keep in mind!

jonasfj:
Mostly consider if there is a better name, probably something in GitHub
actions config will break and have to be patched..

--
Regards Jonas Finnemann Jensen.

tors. 11. jun. 2026, 12.34 skrev Samar Borkar ***@***.***>:

Musharaf-khan:
Hi, I started working on the HMAC primitive part of this issue.

So far I moved the HMAC `importRawKey` dartdoc example into `example/webcrypto/hmac/import_raw_key.dart` and replaced the inline block with:

`{@example /example/webcrypto/hmac/import_raw_key.dart}`

I pushed the branch here: `Apple-beep:webcrypto.dart/docs-hmac-dartdoc-examples`.

I noticed #284 is still open for the initial `example/` restructuring, and `dart analyze` currently reports `doc_directive_unknown` for the new `{@example ...}` directive, which seems related to dart-lang/sdk#63504.

I’ll wait for the initial example-folder PR to land before opening the HMAC primitive PR, unless you prefer I open it now for early feedback.

jonasfj:
Feel free to open WIP PR, just rebase and open a real PR.

jonasfj:
[...]

PR #319, opened 2026-07-13
Enhance documentation for EcdhPublicKey imports

garimaadas22:
This PR adds missing dartdoc examples for `EcdhPublicKey` methods, as part of #283.

**Changes:**
- Added a runnable example for `importRawKey()`
- Added a runnable example for `importSpkiKey()` (existing compatibility notes preserved)
- Added documentation with example for `exportRawKey()`, which previously had no doc comment at all

All examples follow the self-contained pattern (with imports and a runnable `main()`) used elsewhere in the codebase, consistent with #281.

### DIFF
--- a/lib/src/webcrypto/webcrypto.ecdh.dart
+++ b/lib/src/webcrypto/webcrypto.ecdh.dart
@@ -282,8 +282,30 @@ final class EcdhPublicKey {
   factory EcdhPublicKey(EcdhPublicKeyImpl impl) {
     return EcdhPublicKey._(impl);
   }
-
-  /// TODO: find out of this works on Firefox
+/// Import [EcdhPublicKey] from raw elliptic curve point.
+  ///
+  /// Creates an [EcdhPublicKey] from [keyData] given as the raw
+  /// elliptic curve point in uncompressed format. The [curve] specified
+  /// must match the curve used to generate [keyData].
+  ///
+  /// **Example**
+  /// ```dart
+  /// import 'package:webcrypto/webcrypto.dart';
+  ///
+  /// Future<void> main() async {
+  ///   // Generate a key pair.
+  ///   final keyPair = await EcdhPrivateKey.generateKey(EllipticCurve.p256);
+  ///
+  ///   // Export the public key as a raw key.
+  ///   final rawPublicKey = await keyPair.publicKey.exportRawKey();
+  ///
+  ///   // Import the raw key back as an EcdhPublicKey.
+  ///   final publicKey = await EcdhPublicKey.importRawKey(
+  ///     rawPublicKey,
+  ///     EllipticCurve.p256,
+  ///   );
+  /// }
+  /// ```
   static Future<EcdhPublicKey> importRawKey(
     List<int> keyData,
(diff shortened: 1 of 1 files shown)


lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -282,8 +282,30 @@ final class EcdhPublicKey {
   factory EcdhPublicKey(EcdhPublicKeyImpl impl) {
     return EcdhPublicKey._(impl);
   }
-
-  /// TODO: find out of this works on Firefox
+/// Import [EcdhPublicKey] from raw elliptic curve point.
```
jonasfj: dart format!

jonasfj:
See also https://github.com/google/webcrypto.dart/issues/283

mosuem:
LGTM after formatting fix. Moving to the standalone examples from #283 can also be done in a follow-up.

lib/src/webcrypto/webcrypto.ecdh.dart
```diff
@@ -292,6 +315,33 @@ final class EcdhPublicKey {
     return EcdhPublicKey._(impl);
   }
 
+  /// Import [EcdhPublicKey] in the [SPKI][1] format.
+  ///
+  /// Creates an [EcdhPublicKey] from [keyData] given as the DER encoded
```
HamdaanAliQuatil, line 320: nit. could we cite RFC 5480 here as well?

RFC 5280 defines the generic SubjectPublicKeyInfo structure, while RFC 5480 defines the EC algorithm identifiers and point encoding used by this importer. It is also the relevant reference for the id-ecDH / id-ecPublicKey compatibility notes below.

HamdaanAliQuatil (approved):

lib/src/webcrypto/webcrypto.ecdh.dart
garimaadas22: > nit. could we cite RFC 5480 here as well?
> 
> RFC 5280 defines the generic SubjectPublicKeyInfo structure, while RFC 5480 defines the EC algorithm identifiers and point encoding used by this importer. It is also the relevant reference for the id-ecDH / id-ecPublicKey compatibility notes below.

I think it's not mandatory, so we can leave it.

HamdaanAliQuatil: yeah, it isn’t required for correctness. I suggested it for precision because RFC 5480 defines the EC-specific identifiers and encoding discussed in the compatibility notes below, while RFC 5280 defines the generic SPKI structure. 

Since this is non-blocking, I’m good leaving it as-is :)

garimaadas22: > yeah, it isn’t required for correctness. I suggested it for precision because RFC 5480 defines the EC-specific identifiers and encoding discussed in the compatibility notes below, while RFC 5280 defines the generic SPKI structure.
> 
> Since this is non-blocking, I’m good leaving it as-is :)

thanks! Added the RFC 5480 reference.

PR #323, opened 2026-07-14
feat(impl_jni): add JCA RSASSA-PKCS1-v1_5 implementation

mfazrinizar:
## Summary

Adds the JNI/JCA RSASSA-PKCS1-v1_5 implementation on top of the Android JCA backend.

This PR implements:
- RSA key generation with public exponents 3 and 65537
- PKCS#8 private-key and SPKI public-key import/export
- RSA private and public JWK import/export with RS1, RS256, RS384, and RS512 metadata
- strict DER consumption and RSA public/private component validation
- byte and streaming signing and verification through JCA `Signature`
- reusable fixed-size JNI buffers for stream processing
- malformed-signature handling as verification failure
- deterministic JNI reference and throwable cleanup
- focused desktop JNI tests and an Android smoke test

## Known Limitations

- RSA key-generation support remains provider-dependent within the package-wide accepted range. Desktop SunJCE rejects 256-bit generation; the 2048- and 3072-bit generation cases exercised by the shared suite pass.

## Testing

- dart run jni:setup

Verified:
- dart analyze
- dart test test/impl_jni_rsassapkcs1v15_test.dart
- dart test test/webcrypto_test.dart -p vm -n RSASSA-PKCS1-v1_5
- flutter test integration_test/jni_rsassapkcs1v15_test.dart -d emulator-name

### DIFF
--- a/lib/src/impl_jni/impl_jni.rsa_common.dart
+++ b/lib/src/impl_jni/impl_jni.rsa_common.dart
@@ -1,0 +1,438 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+part of 'impl_jni.dart';
+
+typedef _RsaKeyPairData = ({Uint8List privateKeyData, Uint8List publicKeyData});
+
+_HashImpl _rsaHashFromHash(HashImpl hash) {
+  if (hash is _HashImpl) {
+    return hash;
+  }
+  throw AssertionError('Custom implementations of HashImpl are not supported.');
+}
+
+extension _RsaHashMetadata on _HashImpl {
+  String get _rsassaPkcs1v15JcaName {
+    return switch (_jcaName) {
+      'SHA-1' => 'SHA1withRSA',
+      'SHA-256' => 'SHA256withRSA',
+      'SHA-384' => 'SHA384withRSA',
+      'SHA-512' => 'SHA512withRSA',
+      _ => throw AssertionError('Unknown hash algorithm: $_jcaName'),
+    };
+  }
+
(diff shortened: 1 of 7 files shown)


lib/src/impl_jni/impl_jni.rsassapkcs1v15.dart
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC
+// Copyright 2026 Google LLC
```
jonasfj: prefer unrelated changes in separate PRs

mfazrinizar: Noted. Should I amend this? Or just note for next PR?

HamdaanAliQuatil: yes please revert the copyright to 2020. Let's keep copyright-only changes out of this functional PR.

mfazrinizar: Done.

lib/src/impl_jni/impl_jni.rsa_common.dart
```diff
@@ -0,0 +1,643 @@
...
+// limitations under the License.
+
+part of 'impl_jni.dart';
+
+typedef _RsaKeyPairData = ({Uint8List privateKeyData, Uint8List publicKeyData});
+
+// Keep imported-key limits aligned with rsa_check_public_key(), which backs
+// the FFI implementation's RSA_check_key() validation. See:
+// third_party/boringssl/src/crypto/fipsmodule/rsa/rsa_impl.cc.inc
+const _rsaMinModulusBits = 512;
```
HamdaanAliQuatil: blocking: this disagrees with `_validateRsaKeyGenerationParameters` below and with the existing FFI backend, both of which accept 256-bit keys. if a provider does generate a 256-bit key, `generateKey` serializes it and then immediately rejects it during the caller-isolate import as invalid.

please use 256 here so imports and generation preserve the package-wide accepted range. a provider can still reject weak key generation itself.

HamdaanAliQuatil (changes requested):

lib/src/impl_jni/impl_jni.rsa_common.dart
mfazrinizar: Fixed. Import prevalidation now accepts the package-wide 256-bit minimum, matching _validateRsaKeyGenerationParameters, while leaving stronger construction/generation limits to the JCA provider.

Also, updated the nearby comment so it no longer claims exact alignment with BoringSSL's internal minimum.

HamdaanAliQuatil (approved):

PR #324, opened 2026-07-15
refactor: migrate ffigen configs to Dart and upgrade to ^21.0.0

mosuem:
Migrated YAML ffigen configurations to Dart configurations in `tool/ffigen.dart`, updated `ffigen` to `^21.0.0`, updated `tool/update-bindings.sh`, and re-generated native bindings.


jonasfj:
Why not use YAML files?

mosuem:
> Why not use YAML files?

https://pub.dev/packages/ffigen#yaml-configuration-reference It's being deprecated :)

tool/ffigen.dart
```diff
@@ -0,0 +1,353 @@
...
+// ignore_for_file: unused_element
+// ignore_for_file: non_constant_identifier_names
+''',
+    ),
+    headers: Headers(entryPoints: [packageRoot.resolve('src/webcrypto.h')]),
+    functions: Functions(
+      include: Declarations.includeSet({'webcrypto_get_CBB_size'}),
+    ),
+    structs: const Structs(dependencies: CompoundDependencies.opaque),
+    typedefs: const Typedefs(include: Declarations.includeAll),
```
HamdaanAliQuatil: This is pulling in far more than the YAML config did. The regenerated wrapper binding gains 141 typedefs, while the BoringSSL binding gains hundreds of typedefs and more than 100 unrelated opaque declarations. The set of bound functions itself has not changed.

Can we remove includeAll here and in the second config below, and let ffigen generate only the typedefs required by the selected declarations?

example/webcrypto_demo_flutter_app/lib/main.dart
```diff
@@ -24,11 +24,13 @@ void main() {
 }
 
 class MyApp extends StatefulWidget {
+  const MyApp({super.key});
+
   @override
-  _MyAppState createState() => _MyAppState();
+  MyAppState createState() => MyAppState();
```
HamdaanAliQuatil, line 30: Could we leave the demo changes out of this PR? They aren’t needed for the ffigen migration.

If the private return type was causing a lint, the narrower fix is:

```
State<MyApp> createState() => _MyAppState();
```

That keeps the state class private. The scrolling, spacing and deprecated-API cleanup would be easier to review separately.

pubspec.yaml
```diff
@@ -22,13 +22,13 @@ environment:
 
 dependencies:
   ffi: ^2.0.0
-  hooks: ^1.0.0
+  hooks: ^2.0.0
```
HamdaanAliQuatil, line 25: Is the hooks 2.x upgrade required by ffigen 20? This changes a runtime dependency, while the rest of the PR is a code-generation tooling migration.

If it is independent, can we update it separately?

HamdaanAliQuatil (changes requested):
Moving the ffigen configuration to Dart makes sense since YAML configuration is being phased out.

I left comments on three things that currently make this harder to audit: typedefs: includeAll is greatly expanding the generated surface, the Flutter demo cleanup is unrelated, and the hooks major upgrade should be separated unless ffigen 20 requires it.

Could we keep this PR focused on the Dart configuration migration, the ffigen upgrade, the update script, and the necessary regenerated bindings? Changes to generated native bindings are much easier to review when the surrounding diff has one clear purpose.

pubspec.yaml
mosuem: I think its fine to do here IMO.

example/webcrypto_demo_flutter_app/lib/main.dart
mosuem: I just applied `dart fix` as a cleanup.

tool/ffigen.dart
mosuem: done

jonasfj (approved):

HamdaanAliQuatil (approved):

PR #325, opened 2026-07-15
Add artifact fetch build support

mosuem:
## Summary
- Configures `hooks.user_defines.webcrypto.buildMode: fetch` as the default build mode in `pubspec.yaml`.
- Updates `hook/build.dart` to support `fetch` (default), `build` (local CMake compilation), and `local` (custom local binary) modes.
- Adds SHA256 integrity checks against `lib/src/hook_helpers/hashes.dart`.
- Adds `tool/regenerate_hashes.dart` to autogenerate release hash tables.
- Adds `.github/workflows/release-artifacts.yml` covering Linux, macOS (x64, arm64), Windows, Android (arm64, arm, x64), and iOS (arm64).
- Enables manual `workflow_dispatch` for workflow testing without publishing releases.
- Pins GitHub Actions to full commit SHA hashes.

After merging, we need to run the artifact generation once to get the hashes.


HamdaanAliQuatil:
Thanks for putting this together. ICU4X is a useful reference, and I like the checksum verification and the explicit build/local modes.

I think we should choose the distribution model before getting too far into the implementation. #315 includes the CI-build binaries in the pub package, while this PR downloads only the required target binary from GitHub Releases.

The fetch approach has a good size advantage once we add mobile targets and static archives. The packaged approach makes releases and offline builds simpler. I don’t think runtime fetching is necessarily a blocker, but we should agree on how assets and hashes are produced before publishing, what happens when GitHub cannot be reached, and whether release assets are treated as immutable.

There are also a few implementation details to settle, including the current 0.6.1 versus v0.6.1 tag naming and validation of the Android and iOS outputs. I’d merge #324 separately so this PR stays focused on artifact delivery.

@jonasfj, which direction do you think makes more sense for us long-term? It may also be worth comparing the actual total artifact size before choosing.

cornpip:
Tried this PR out on Windows while evaluating it for a downstream package — ran into one issue:

`buildMode: local` with an absolute Windows `localPath` fails — the drive letter is parsed as a URI scheme.

Repro branch [link](https://github.com/cornpip/litert_crypto/tree/repro/webcrypto-325-localpath) — pins this PR's head, `b2127e3a`:

```sh
git clone -b repro/webcrypto-325-localpath https://github.com/cornpip/litert_crypto
cd litert_crypto/example
flutter pub get
dart run litert_crypto encrypt
```

The pubspec sets:

```yaml
dependency_overrides:
  webcrypto:
    git:
      url: https://github.com/mosuem/webcrypto.dart
      ref: b2127e3afe74c00d22b4a84e7bc7ed785ae39a37

hooks:
  user_defines:
    webcrypto:
      buildMode: local
      localPath: D:\any\path\webcrypto.dll
```

and the hook dies with:

```text
Running build hooks... Unhandled exception:
Unsupported operation: Cannot extract a file path from a d URI
#0      _Uri.toFilePath (dart:core/uri.dart:3432:7)
#1      new File.fromUri (dart:io/file.dart:212:45)
#2      _useLocalBinary (file:///d:/pubcache/git/webcrypto.dart-b2127e3afe74c00d22b4a84e7bc7ed785ae39a37/hook/build.dart:156:21)

  Building assets for package:webcrypto failed.
  build.dart returned with exit code: 255.
  stdout:
  webcrypto: build options: BuildOptions(buildMode: BuildModeEnum.local, localPath: d:/any/path/webcrypto.dll)
```

`D:\...` becomes a URI with scheme `d` (visible in the hook's own log line), so `File.fromUri` at `hook/build.dart:156` throws — before the existence check, so the path doesn't even need to exist.

A relative `localPath` works fine (resolves against the pubspec to a proper `file:///D:/...`), so this is only about how `defines.path('localPath')` handles Windows absolute paths.

Tested on Windows 11 x64 with Flutter 3.44.8 (stable) / Dart 3.12.2.

Issue #263, opened 2026-04-30
Apply coverage ignore comments as applicable


jonasfj:
From: https://pub.dev/packages/coverage

> **Ignore lines from coverage**
> * `// coverage:ignore-line` to ignore one line.
> * `// coverage:ignore-start` and // coverage:ignore-end to ignore range of lines inclusive.
> * `// coverage:ignore-file` to ignore the whole file.

We should apply this to bit of the code that we don't want coverage for.
I'm sure we have a few places that we don't care to test.

**Example**, maybe it's overkill to test `requireSubtleCrypto` logic, we could just make that private method, and use these coverage-ignore comments to avoid counting it in coverage.
https://github.com/google/webcrypto.dart/blob/7a673ab4bdf2c4638e819aa75608e019a3488fb7/lib/src/crypto_subtle.dart#L81-L102

This is really just nice to have and generally users shouldn't hit this. And we don't really want to make testing setups where we can reproduce the behavior. So any testing we do is kind of artificial anyways.

AquibAquil:
If this is open could i pick this up?

HamdaanAliQuatil:
hey @AquibAquil. Please go ahead and create a PR. Thanks for contributing!

harrshita123:
@jonasfj Is this issue still open? I would like to pick it up.

PR #327, opened 2026-07-16
chore: mark unsupported code paths with coverage:ignore

AquibAquil:
Addresses part of #263.

## What

Adds `// coverage:ignore-*` markers to code paths documented in-source as intentionally unsupported or unreachable:

**Whole-file ignores — `lib/src/impl_stub/` (14 files)**

Marked with `// coverage:ignore-file`. This directory is loaded only via conditional import when *neither* `dart.library.ffi` nor `dart.library.js_interop` is available (see `lib/src/webcrypto/webcrypto.dart:27`). No real Dart runtime satisfies that condition, so every `throw UnimplementedError('Not implemented')` in the directory is unreachable at runtime by design.

**Block ignores — 4 files in `impl_ffi/`, `impl_js/`, and `boringssl/`**

Wrapped with `// coverage:ignore-start` / `// coverage:ignore-end`:

- Three `throw UnsupportedError('192-bit AES keys are not supported')` sites in `lib/src/impl_ffi/impl_ffi.aes_common.dart` (documented in-source as intentionally unsupported, see https://crbug.com/533699).
- The unsupported-platform fallback throw in `lib/src/boringssl/lookup/utils.dart` (fires only on operating systems outside {Windows, Linux, macOS}).
- Five "This should never happen!" curve fallbacks in `lib/src/impl_ffi/impl_ffi.ec_common.dart` (×3), `lib/src/impl_ffi/impl_ffi.ecdsa.dart`, and `lib/src/impl_js/impl_js.utils.dart` — dead-code fallbacks after exhaustive enum switches.

## Why

Per #263, these are candidates where "we don't want coverage" — code paths intentionally chosen not to test rather than defensively cover.

## Skipped intentionally

- `requireSubtleCrypto` (`lib/src/crypto_subtle.dart:82`) — the example cited in the issue. **Now has tests** at `test/crypto_subtle_test.dart:128` and `:141`, so ignore markers would silence real coverage. Left alone.
- `throw UnimplementedError('Only supports 65537 and 3 for now')` (`lib/src/crypto_subtle.dart:44`) — genuine TODO for future work, not an intentional gap. Left visible.

## Verification

- `dart format --output none --set-exit-if-changed .` passes.
- `dart analyze --fatal-warnings .` passes.
- Grep confirms no existing tests exercise the ignored throws (except `requireSubtleCrypto`, which is why it's excluded).
- Local test run was not possible on Windows: the native asset build fails with `CMake Error at CMakeLists.txt:19 (project): Generator Ninja does not support platform specification, but platform x64 was specified.` Working around it locally was out of scope for this change; relying on CI to validate.

### DIFF
--- a/lib/src/boringssl/lookup/utils.dart
+++ b/lib/src/boringssl/lookup/utils.dart
@@ -29,10 +29,12 @@ String get libraryFileName {
   if (Platform.isMacOS) {
     return 'lib$libraryName.dylib';
   }
+  // coverage:ignore-start
   throw UnsupportedError(
     'Platform ${Platform.operatingSystem} is unsupported or embed '
     'the binary webcrypto library for package:webcrypto',
   );
+  // coverage:ignore-end
 }
 
 /// Look for the webcrypto binary library in the `.dart_tool/webcrypto/` folder.
--- a/lib/src/impl_ffi/impl_ffi.aes_common.dart
+++ b/lib/src/impl_ffi/impl_ffi.aes_common.dart
@@ -18,7 +18,9 @@ Uint8List _aesImportRawKey(List<int> keyData) {
   if (keyData.length == 24) {
     // 192-bit AES is intentionally unsupported, see https://crbug.com/533699
     // If not supported in Chrome, there is not reason to support it in Dart.
+    // coverage:ignore-start
     throw UnsupportedError('192-bit AES keys are not supported');
+    // coverage:ignore-end
   }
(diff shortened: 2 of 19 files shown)


mosuem (changes requested):
After rebasing onto main, this LGTM!

AquibAquil:
I've rebased it. please review.

mosuem (approved):
Perfect, thanks!

PR #329, opened 2026-07-16
fix(ffi): validate JWK base64url encoding

### LINKED ISSUE #328: bug: FFI backend's JWK imports accept padded and standard Base64 values
## Summary

The FFI backend accepts JWK properties that are not valid unpadded base64url.

For example, both of these are currently accepted:
- `AQ==`, which contains padding
- `+w`, which uses the standard Base64 `+` character

Browser WebCrypto rejects both values. JWK fields must use the URL-safe Base64 alphabet without trailing padding, as specified by [RFC 7515](https://www.rfc-editor.org/info/rfc7515/#section-2).

## Reproduction

```dart
await HmacSecretKey.importJsonWebKey({
  'kty': 'oct',
  'alg': 'HS256',
  'k': 'AQ==',
}, Hash.sha256);

await HmacSecretKey.importJsonWebKey({
  'kty': 'oct',
  'alg': 'HS256',
  'k': '+w',
}, Hash.sha256);
```

## Expected Behavior

Both imports throw FormatException consistently across FFI and browser backends.

## Actual Behavior

The FFI backend imports both keys, while the browser backend rejects them. The issue is in the shared FFI JWK decoder, so it can affect HMAC, AES, EC, and RSA JWK fields.

mfazrinizar:
## Summary

Fixes #328.

The shared FFI JWK decoder currently accepts padded values and characters from the standard Base64 alphabet.

This PR validates the decoded value by re-encoding it as canonical unpadded base64url and comparing it with the original input. It therefore rejects:
- explicit Base64 padding
- the standard Base64 alphabet
- other non-canonical encodings

The regression tests use HMAC JWK import as a representative public API path. The decoder is shared by the FFI HMAC, AES, EC, and RSA implementations.

## Testing

Verified with:
- dart format --output=none --set-exit-if-changed .
- git diff --check
- dart analyze
- dart test test/webcrypto_test.dart -p vm -n 'JWK: rejects'
- dart test test/webcrypto_test.dart -p chrome -n 'JWK: rejects'
- dart test test/webcrypto_test.dart -p vm
- dart test test/webcrypto_test.dart -p chrome

### DIFF
--- a/lib/src/impl_ffi/impl_ffi.utils.dart
+++ b/lib/src/impl_ffi/impl_ffi.utils.dart
@@ -486,7 +486,11 @@ Uint8List _jwkDecodeBase64UrlNoPadding(String unpadded, String prop) {
       unpadded.length + ((4 - (unpadded.length % 4)) % 4),
       '=',
     );
-    return base64Url.decode(padded);
+    final decoded = base64Url.decode(padded);
+    if (_jwkEncodeBase64UrlNoPadding(decoded) != unpadded) {
+      throw const FormatException();
+    }
+    return decoded;
   } on FormatException {
     throw FormatException(
       'JWK property "$prop" is not url-safe base64 without padding',
--- a/lib/src/testing/regression/jwk_base64url.dart
+++ b/lib/src/testing/regression/jwk_base64url.dart
@@ -1,0 +1,42 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:webcrypto/webcrypto.dart';
+
+import '../utils/utils.dart';
+
+List<({String name, Future<void> Function() test})> tests() => [
+  (
+    name: 'JWK: rejects padded base64url key data',
+    test: () => _checkInvalidKeyData('AQ=='),
+  ),
+  (
+    name: 'JWK: rejects the standard base64 alphabet',
+    test: () => _checkInvalidKeyData('+w'),
+  ),
+];
+
+Future<void> _checkInvalidKeyData(String keyData) async {
+  var threw = false;
+  try {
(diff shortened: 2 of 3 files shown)


mosuem:
I am a bit torn, this seems to be quite a heavy way to do the check - re-encoding does more work than is actually necessary. But then again we can always make it faster later...

mfazrinizar:
> I am a bit torn, this seems to be quite a heavy way to do the check - re-encoding does more work than is actually necessary. But then again we can always make it faster later...

I think that’s fair. The round-trip check was considered because it makes canonical unpadded base64url validation simple and difficult to get wrong, but it does add another encoding pass, allocation, and comparison.

Maybe a more efficient version would validate the URL-safe alphabet, length, and unused final bits directly in one pass, then add padding and decode once. Since this only runs during JWK import and the inputs are relatively small, I agree it is not urgent, but we can replace it with the direct validator later if this becomes worth optimizing.

HamdaanAliQuatil:
agreed that a direct validator would be leaner. for this correctness fix, the round-trip check is easier to audit and JWK import is not on the normal crypto-operation path. we can replace it if a concrete import-heavy workload makes the extra allocation measurable.

HamdaanAliQuatil (approved):

lib/src/testing/regression/jwk_base64url.dart
```diff
@@ -0,0 +1,42 @@
...
+import '../utils/utils.dart';
+
+List<({String name, Future<void> Function() test})> tests() => [
+  (
+    name: 'JWK: rejects padded base64url key data',
+    test: () => _checkInvalidKeyData('AQ=='),
+  ),
+  (
+    name: 'JWK: rejects the standard base64 alphabet',
+    test: () => _checkInvalidKeyData('+w'),
```
jonasfj, line 26: is + not part of url-safe base64?

lib/src/impl_ffi/impl_ffi.utils.dart
```diff
@@ -486,7 +486,11 @@ Uint8List _jwkDecodeBase64UrlNoPadding(String unpadded, String prop) {
       unpadded.length + ((4 - (unpadded.length % 4)) % 4),
       '=',
     );
-    return base64Url.decode(padded);
+    final decoded = base64Url.decode(padded);
+    if (_jwkEncodeBase64UrlNoPadding(decoded) != unpadded) {
+      throw const FormatException();
```
jonasfj, line 491: This is a bit messy, why not make a RegExp and test with this upfront?

If people are debugging this and break on all exceptions they'll first see a empty format exception and then one with a message int.

Would a regex be much harder to do?

jonasfj:
Okay, I see this might be the most reliable way to do it, and easiest. Feel free to file an issue to make a better implementation, I'll land this for now.

PR #330, opened 2026-07-16
docs: correct JWK use value for AES keys

w1boost1889M:
## Summary

Correct the documented JWK `use` value for AES-CBC, AES-CTR, and AES-GCM keys.

The documentation currently states that the `use` property must be `"sig"`, but AES keys are encryption keys and the implementation requires `"enc"` when the property is present.

## Changes

* Correct `AesCbcSecretKey.importJsonWebKey` documentation.
* Correct `AesCtrSecretKey.importJsonWebKey` documentation.
* Correct `AesGcmSecretKey.importJsonWebKey` documentation.

## Testing

Documentation-only change. No runtime behavior is modified.


mosuem (approved):
Thanks!

PR #333, opened 2026-07-20
fix(ffi): reject private JWKs in RSA public imports

### LINKED ISSUE #332: bug: native RSA public JWK imports accept private keys
## Summary

The native FFI backend accepts an RSA private JWK when it is passed to an RSA public-key importer. The import succeeds after silently discarding the private parameters (`d`, `p`, `q`, `dp`, `dq`, and `qi`).

This differs from Web Crypto implementations, which reject importing a private RSA JWK with public-key usages.

The shared behavior affects:

- `RsaOaepPublicKey.importJsonWebKey`
- `RsaPssPublicKey.importJsonWebKey`
- `RsassaPkcs1V15PublicKey.importJsonWebKey`

## Reproduction

```dart
final pair = await RsaOaepPrivateKey.generateKey(
  1024,
  BigInt.from(65537),
  Hash.sha256,
);

final privateJwk = await pair.privateKey.exportJsonWebKey();

final publicKey = await RsaOaepPublicKey.importJsonWebKey(
  privateJwk,
  Hash.sha256,
);

print(publicKey); 
```

## Actual behavior

The native backend imports `n` and `e`, ignores all private parameters, and returns an `RsaOaepPublicKey`.

The same behavior is present in the RSA-PSS and RSASSA-PKCS1-v1_5 public-key import paths because they use the shared RSA JWK importer.

## Expected behavior

The import should reject the JWK because the presence of `d` identifies it as a private RSA key. Standards-based Web Crypto implementations reject the same import because a private RSA key cannot be assigned public-key usages such as `encrypt` or `verify`.

The EC JWK importer already enforces the equivalent invariant by rejecting `d` when importing a public key.

## Cause

`_importJwkRsaPrivateOrPublicKey` validates that `d` is present for private-key imports, but it does not validate that `d` is absent for public-key imports.

## Proposed fix

Reject RSA JWKs containing `d` when `isPrivateKey` is false in the shared RSA import path, and add regression coverage for RSA-OAEP, RSA-PSS, and RSASSA-PKCS1-v1_5 public-key imports.

harrshita123:
## Summary

- Reject RSA JWKs containing the private exponent when importing a public key through the native FFI backend.
- Add regression coverage for RSA-OAEP, RSA-PSS, and RSASSA-PKCS1-v1_5 public-key imports.

## Root cause

The shared native RSA JWK importer required `d` for private-key imports but did not require it to be absent for public-key imports. Public-key import methods therefore accepted private JWKs and silently discarded their private parameters.

The new validation rejects the JWK before allocating or parsing the native RSA key.

## Validation

- `dart analyze`
- `dart test test/rsa_jwk_validation_test.dart -p vm`
- `dart test test/rsa_jwk_validation_test.dart -p chrome`
- `dart test -p vm`

Fixes #332

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # 0.6.2-wip
 * Fixed JS interop to enable WebAssembly.
+* Fixed native RSA public-key JWK imports to reject private key material.
 
 # 0.6.1
 * Added Dart native build hooks and native asset lookup for the bundled
--- a/lib/src/impl_ffi/impl_ffi.rsa_common.dart
+++ b/lib/src/impl_ffi/impl_ffi.rsa_common.dart
@@ -86,6 +86,9 @@ _EvpPKey _importJwkRsaPrivateOrPublicKey(
       'use',
       'must be "$expectedUse", if present',
     );
+    if (!isPrivateKey) {
+      checkJwk(jwk.d == null, 'd', 'must not be present for a public key');
+    }
 
     // TODO: Consider rejecting private keys with 'use', as it's only valid for
     //       public keys according to the RFC -- maybe read the RFC again to be
--- a/test/rsa_jwk_validation_test.dart
+++ b/test/rsa_jwk_validation_test.dart
@@ -1,0 +1,62 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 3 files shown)


HamdaanAliQuatil (approved):
This looks correct. An RSA JWK with d is a private key, so the public import APIs should reject it instead of silently using only n and e. Keeping the check in the shared RSA importer covers OAEP, PSS, and PKCS#1 and brings the native backend in line with Web Crypto.

I ran the focused test on VM and Chromium, the full VM suite, and dart analyze locally; all passed. Could you also add a short entry under 0.6.2-wip in the changelog?

harrshita123:
> This looks correct. An RSA JWK with d is a private key, so the public import APIs should reject it instead of silently using only n and e. Keeping the check in the shared RSA importer covers OAEP, PSS, and PKCS#1 and brings the native backend in line with Web Crypto.
> 
> I ran the focused test on VM and Chromium, the full VM suite, and dart analyze locally; all passed. Could you also add a short entry under 0.6.2-wip in the changelog?

I have added a short entry under 0.6.2-wip in the changelog to document the RSA public JWK import fix.

HamdaanAliQuatil (approved):
lgtm :)

PR #334, opened 2026-07-20
Build native libraries with Dart Native Assets

jpohhhh:
## Summary

Replace the five native Flutter-plugin build wrappers and manual test setup path with one Dart Native Assets build hook for Android, iOS, Linux, macOS, and Windows.

The package now:

- builds the existing pinned vendored BoringSSL source and symbol-table wrapper from one canonical `src/CMakeLists.txt`;
- consumes the Native Assets target compiler/toolchain/SDK configuration through `native_toolchain_cmake`;
- registers one `DynamicLoadingBundled` code asset with the lookup library as its package identity;
- resolves the asset through `@Native`, with narrow bundle-root and legacy direct-script fallbacks;
- caches expensive native builds by source/toolchain/target content under a process-safe lock and atomically publishes a SHA-256-validated result;
- keeps only the browser Flutter plugin declaration and removes Android, Darwin, Linux, and Windows native plugin wrappers;
- leaves `webcrypto:setup` as a documented legacy fallback rather than a prerequisite for supported tests/scripts;
- upgrades the old desktop example templates to install generic package Native Assets;
- adds final-artifact inspection and no-setup runtime coverage across all supported targets.

No API, algorithm, wire/key format, generated symbol index, or BoringSSL source revision is changed.

## Validation

Final CI: https://github.com/Telosnex/webcrypto.dart/actions/runs/29781648591

Coverage includes:

- `dart test` and `flutter test` on Linux, macOS, and Windows without `webcrypto:setup`;
- Linux, macOS, Windows, Android-emulator, and iOS-simulator integration runtime tests;
- Android release APK inspection for all emitted ABIs;
- unsigned iOS ARM64 app inspection;
- universal macOS app inspection;
- Linux ELF and Windows PE/DLL inspection, including `webcrypto_lookup_symbol` export checks;
- browser JS and Wasm tests plus rejection of native files in Web output;
- concurrent cold-cache builds, cache digest integrity, and source-mutation invalidation;
- `flutter pub publish --dry-run` checks for hook/CMake/vendored inputs.

The same package hook was also cross-built on an Apple Silicon macOS host for Linux ARM64 and run under the bare `flutter-pi` embedder on Raspberry Pi 5. The generated AArch64 ELF was loaded from the application bundle without `LD_LIBRARY_PATH`; a real persisted E2EE account downloaded its wrapped master-key blob, verified its canary, passed its health check, and survived service restarts.

## Migration notes

- Minimum SDK is Dart 3.10 / Flutter 3.38, where package build hooks are supported by the advertised workflows.
- Native consumers need the platform's normal C/C++ toolchain and CMake. Windows x86/x64 also requires NASM.
- Applications generated from older Flutter desktop templates may need Flutter's standard Native Assets CMake migration. The example in this repository now includes the current generic install stanza.
- Browser builds remain backed by the browser Web Crypto API and produce no BoringSSL asset.


jpohhhh:
Closing because current upstream master has independently landed Native Assets/plugin removal and subsequently moved to direct FFI bindings with a newer BoringSSL tree. This fork branch intentionally ports all-platform Native Assets to Telosnex's older pinned baseline; merging it now would overwrite newer upstream crypto/build work rather than provide a safe incremental change. The implementation remains maintained and tested in the Telosnex fork.

PR #335, opened 2026-07-21
chore(jca): add RSA parameter JNI bindings

mfazrinizar:
## Summary

Adds JNIgen coverage for the parameter classes required by the JCA RSA-PSS and
RSA-OAEP implementations:
- `java.security.spec.MGF1ParameterSpec`
- `java.security.spec.PSSParameterSpec`
- `javax.crypto.spec.OAEPParameterSpec`
- `javax.crypto.spec.PSource` and its generated `PSource.PSpecified` binding

The first commit updates the reproducible generation inputs. The second commit contains only regenerated output. 

## Regeneration

- bash tool/update-jca-bindings.sh

The generator remains pinned to Android SDK Platform 36 and JDK 17. Running the generator a second time produces the same generated output and no additional diff.

## Testing

- dart format --output=none --set-exit-if-changed .
- dart analyze
- git diff --check

### DIFF
--- a/lib/src/third_party/jca/generated_bindings.dart
+++ b/lib/src/third_party/jca/generated_bindings.dart
@@ -6471,6 +6471,536 @@ final class $PKCS8EncodedKeySpec$Type$
   String get signature => r'Ljava/security/spec/PKCS8EncodedKeySpec;';
 }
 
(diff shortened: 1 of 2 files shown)


HamdaanAliQuatil (approved):
looks good

PR #336, opened 2026-07-21
Upgrade native_toolchain_cmake to v0.3

emersion:
This new version improves Android SDK detection: https://github.com/rainyl/native_toolchain_cmake/issues/31


HamdaanAliQuatil (approved):
thanks :) we should update the example lockfile too.

PR #337, opened 2026-07-23
fix: use canonical RSA-OAEP JWK alg for SHA-1

mfazrinizar:
## Summary

Uses the canonical `RSA-OAEP` JWK `alg` identifier for RSA-OAEP with SHA-1 while preserving compatibility with JWKs exported by older FFI releases.

This PR:
- changes FFI SHA-1 RSA-OAEP exports from `RSA-OAEP-1` to `RSA-OAEP`
- imports canonical `RSA-OAEP` private and public JWKs with SHA-1
- accepts the historical `RSA-OAEP-1` value only when importing with SHA-1
- rejects `RSA-OAEP-1` when importing with SHA-256, SHA-384, or SHA-512
- avoids mutating caller-provided JWK maps
- adds reusable public API regression coverage across native and browser backends

## Testing

- `dart analyze`
- `dart test test/webcrypto_test.dart -p vm -n RSA-OAEP`
- `dart test test/webcrypto_test.dart -p chrome -c dart2js -n RSA-OAEP`
- `dart test test/webcrypto_test.dart -p chrome -c dart2wasm -n RSA-OAEP`
- `dart test test/webcrypto_test.dart -p vm`
- `dart test test/webcrypto_test.dart -p chrome -c dart2js`
- `dart test test/webcrypto_test.dart -p chrome -c dart2wasm`

### DIFF
--- a/lib/src/webcrypto/webcrypto.rsaoaep.dart
+++ b/lib/src/webcrypto/webcrypto.rsaoaep.dart
@@ -14,6 +14,17 @@
 
 part of 'webcrypto.dart';
 
+Map<String, dynamic> _normalizeLegacyRsaOaepJwkAlg(
+  Map<String, dynamic> jwk,
+  Hash hash,
+) {
+  // Preserve imports of JWKs exported by older FFI releases.
+  if (identical(hash, Hash.sha1) && jwk['alg'] == 'RSA-OAEP-1') {
+    return {...jwk, 'alg': 'RSA-OAEP'};
+  }
+  return jwk;
+}
+
 /// RSAES-OAEP private key for decryption of messages.
 ///
 /// An [RsaOaepPrivateKey] instance holds a private RSA key for decrypting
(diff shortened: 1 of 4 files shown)


HamdaanAliQuatil (approved):

lib/src/webcrypto/webcrypto.rsaoaep.dart
```diff
@@ -14,6 +14,17 @@
 
 part of 'webcrypto.dart';
 
+Map<String, dynamic> _normalizeLegacyRsaOaepJwkAlg(
+  Map<String, dynamic> jwk,
+  Hash hash,
+) {
+  // Preserve imports of JWKs exported by older FFI releases.
+  if (identical(hash, Hash.sha1) && jwk['alg'] == 'RSA-OAEP-1') {
```
jonasfj, line 22: Why use `identical` is there a operator override here?

But also I could be convinced to not do this, and just accept that some JWKs we made in the past don't work anymore.

We are not at 1.0.0 yet, and honestly, these JWKs are just broken.

btw, this fix is worth noting in the changelog.

PR #339, opened 2026-07-24
docs: correct RSA-OAEP plaintext size limit

mfazrinizar:
## Summary

Corrects the documented RSA-OAEP plaintext size limit.

The previous formula mixed bit and byte units and stated that a 2048-bit RSA key with SHA-256 could encrypt 191 bytes. The PKCS #1 limit is: `modulusByteLength - 2 * hashByteLength - 2`

For a 2048-bit key with SHA-256, the maximum plaintext size is 190 bytes.

This updates both the reusable RSA-OAEP message-size documentation and the `encryptBytes` documentation. There are no behavior changes.

## Testing

- `dart analyze`
- `dart doc --dry-run`
- `dart test test/webcrypto_test.dart -p vm -n RSA-OAEP`
- `git diff --check`


HamdaanAliQuatil (approved):
lgtm :)

PR #341, opened 2026-07-24
feat(impl_jni): add JCA RSA-PSS implementation

mfazrinizar:
## Summary

Adds the JNI/JCA RSA-PSS implementation on top of the Android JCA backend and the persistent RSA key ownership introduced in #323.

This PR implements:
- RSA key generation with public exponents 3 and 65537
- PKCS#8 private-key and SPKI public-key import/export
- RSA-PSS JWK import/export
- byte and stream signing and verification
- explicit digest, MGF1 digest, salt-length, and trailer-field parameters
- standard `RSASSA-PSS` provider lookup with hash-specific Android aliases
- focused desktop JNI tests and an Android smoke test covering all supported hashes

Persistent JCA keys are retained through the shared isolate-unsendable key
owner. JNI arenas are used only for temporary values created by an operation.

## Testing

```sh
dart run jni:setup
```

Verified:
```sh
dart analyze
dart test test/impl_jni_rsapss_test.dart
dart test test/webcrypto_test.dart -p vm -n RSA-PSS
flutter test integration_test/jni_rsapss_test.dart -d emulator-name
git diff --check
```

### DIFF
--- a/lib/src/impl_jni/impl_jni.rsapss.dart
+++ b/lib/src/impl_jni/impl_jni.rsapss.dart
@@ -244,6 +244,12 @@ Signature _createRsaPssSignature(
   return signature;
 }
 
+// TODO: Decide the RSA-PSS policy for Android API 21-22. Android only
+// guarantees the hash-specific SHAxxxwithRSA/PSS aliases from API 23, while
+// RSASSA-PSS availability depends on the installed provider. Possible policies
+// include documenting provider-dependent support, requiring API 23, or routing
+// unsupported cases to another backend. See:
+// https://developer.android.com/reference/java/security/Signature
 Signature _getRsaPssSignature(jni.Arena arena, _HashImpl hash) {
   try {
     return _getRsaPssSignatureByName(arena, 'RSASSA-PSS');
--- a/lib/src/impl_jni/impl_jni.rsapss.dart
+++ b/lib/src/impl_jni/impl_jni.rsapss.dart
@@ -270,7 +270,7 @@ void _validateRsaPssSaltLength(int saltLength) {
     throw ArgumentError.value(
       saltLength,
       'saltLength',
-      'must be a positive integer',
+      'must be a non-negative integer',
     );
   }
 }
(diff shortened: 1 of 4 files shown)


lib/src/impl_jni/impl_jni.rsapss.dart
```diff
@@ -21,32 +21,278 @@ final class _StaticRsaPssPrivateKeyImpl implements StaticRsaPssPrivateKeyImpl {
...
+  return signature;
+}
+
+void _validateRsaPssSaltLength(int saltLength) {
+  if (saltLength < 0) {
+    throw ArgumentError.value(
+      saltLength,
+      'saltLength',
+      'must be a positive integer',
+    );
```
HamdaanAliQuatil, line 280: nit: zero is a valid rsa-pss salt length, including in our shared vectors, so this should say `must be a non-negative integer`

HamdaanAliQuatil (approved):

HamdaanAliQuatil:
non-blocking follow-up: please record RSA-PSS as a provider/API-level limitation. Android’s hash-specific SHAxxxwithRSA/PSS aliases begin at API 23, while the generic RSASSA-PSS entry is not available on API 9-22. so API 21-22 support is provider-dependent and cannot be guaranteed by this implementation alone.

see https://developer.android.com/reference/java/security/Signature

lib/src/impl_jni/impl_jni.rsapss.dart
mfazrinizar: Thanks. Updated the message to `must be a non-negative integer`.

mfazrinizar:
> non-blocking follow-up: please record RSA-PSS as a provider/API-level limitation. Android’s hash-specific SHAxxxwithRSA/PSS aliases begin at API 23, while the generic RSASSA-PSS entry is not available on API 9-22. so API 21-22 support is provider-dependent and cannot be guaranteed by this implementation alone.
> 
> see https://developer.android.com/reference/java/security/Signature

Noted. Recorded this as an inline TODO next to the RSA-PSS provider selection. It documents API 21–22 as provider-dependent and leaves the eventual policy open: document limited support, require API 23, or route unsupported cases through another backend.

HamdaanAliQuatil (approved):

PR #343, opened 2026-07-25
fix(hmac): use block size for default key length

### LINKED ISSUE #342: bug: native HMAC generation uses digest size instead of hash block size by default
### Description

`HmacSecretKey.generateKey(hash)` produces different default key lengths on native and browser backends.

When `length` is omitted, Web Crypto requires the recommended HMAC key length to be the block size of the selected hash function. The browser backend delegates this choice to `SubtleCrypto.generateKey` and follows that rule, but the FFI backend uses `EVP_MD_size`, which is the digest size rather than the block size.

This gives the following results:

| Hash | FFI backend | Web Crypto / browser |
| --- | ---: | ---: |
| SHA-1 | 160 bits | 512 bits |
| SHA-256 | 256 bits | 512 bits |
| SHA-384 | 384 bits | 1024 bits |
| SHA-512 | 512 bits | 1024 bits |

The public documentation currently describes the FFI behavior, so it also differs from Web Crypto's definition.

### Reproduction

```dart

Future<void> main() async {
  for (final entry in <String, Hash>{
    'SHA-1': Hash.sha1,
    'SHA-256': Hash.sha256,
    'SHA-384': Hash.sha384,
    'SHA-512': Hash.sha512,
  }.entries) {
    final key = await HmacSecretKey.generateKey(entry.value);
    final raw = await key.exportRawKey();
    print('${entry.key}: ${raw.length * 8} bits');
  }
}
```

On the FFI backend this prints:

```text
SHA-1: 160 bits
SHA-256: 256 bits
SHA-384: 384 bits
SHA-512: 512 bits
```

### Expected behavior

Omitting `length` should generate a key whose length equals the hash function's block size:

- 512 bits for SHA-1 and SHA-256
- 1024 bits for SHA-384 and SHA-512

An explicitly supplied `length` should continue to take precedence unchanged.

This matches the HMAC key generation algorithm in the Web Cryptography specification:
https://www.w3.org/TR/WebCryptoAPI/#hmac-operations

### Actual behavior

The FFI implementation defaults to:

```dart
length ??= ssl.EVP_MD_size(h._md) * 8;
```

`EVP_MD_size` returns the digest size, causing native-generated keys to be shorter than browser-generated keys for every supported hash.

### Suggested fix

- Represent the HMAC default block size for each supported hash in the native implementation.
- Use it only when `length` is omitted.
- Correct the `HmacSecretKey.generateKey` documentation.
- Add regression coverage for all supported hashes and for an explicit length.

harrshita123:
## Summary

- use each hash function's block size as the default native HMAC key length
- keep omitted lengths consistent across backend adapters
- document the Web Crypto default and distinguish block size from digest size
- run default-length regression coverage through the shared native and browser test suite

## Root cause

The FFI backend used `EVP_MD_size` when `length` was omitted. That function returns the digest size, while Web Crypto defines the default HMAC key length as the hash function's block size. This made native-generated keys shorter than browser-generated keys.

Explicitly supplied key lengths continue to take precedence.

## Validation

- `dart format --output none --set-exit-if-changed .`
- focused shared regression on the VM
- focused shared regression on Chrome with dart2js and dart2wasm
- full VM suite: 1,462 passed, 1 platform-specific skip

Closes #342


lib/src/webcrypto/webcrypto.hmac.dart
```diff
@@ -181,8 +181,9 @@ final class HmacSecretKey {
   /// Generate random [HmacSecretKey].
   ///
   /// The [length] specifies the length of the secret key in bits. If omitted
-  /// the random key will use the same number of bits as the underlying hash
-  /// algorithm given in [hash].
+  /// the random key will use the block size of the underlying hash algorithm
+  /// given in [hash] (512 bits for SHA-1 and SHA-256, and 1024 bits for
```
jonasfj: this is not true, sha1 is 160 bits, sha-256 is not 512 bits.

test/hmac_default_key_length_test.dart
jonasfj, line 1: better use lib/src/testing/regression/ why not?

test/hmac_default_key_length_test.dart
```diff
@@ -0,0 +1,39 @@
...
+import 'package:test/test.dart';
+import 'package:webcrypto/webcrypto.dart';
+
+void main() {
+  group('HMAC default key length', () {
+    for (final testCase in [
+      (name: 'SHA-1', hash: Hash.sha1, expectedBits: 512),
+      (name: 'SHA-256', hash: Hash.sha256, expectedBits: 512),
+      (name: 'SHA-384', hash: Hash.sha384, expectedBits: 1024),
+      (name: 'SHA-512', hash: Hash.sha512, expectedBits: 1024),
```
jonasfj: Are we sure this what browsers do?

Does the spec say anything?

I could be convinced to now allow a default value, and that we always require a length be specificed.

jonasfj (changes requested):

lib/src/webcrypto/webcrypto.hmac.dart
harrshita123: Thanks for pointing out that the wording was confusing. I updated it to describe Web Crypto's recommended HMAC key length and explicitly distinguish the hash function's block size from its digest size. I also linked the HMAC generation section of the specification.

test/hmac_default_key_length_test.dart
harrshita123: I moved these checks into `lib/src/testing/regression/hmac_default_key_length.dart` and registered them with `runAllTests`, so they now run through the shared native and browser test harness.

test/hmac_default_key_length_test.dart
harrshita123: Yes. Web Crypto section 31.6.3 specifies that an omitted HMAC length uses the hash function's block size in bits. Chromium follows the same rule through `GetDigestBlockSizeBits`. I kept the parameter optional to match the specification and avoid a breaking API change. The shared regression passes on the VM, Chrome with dart2js, and Chrome with dart2wasm.

Spec: https://www.w3.org/TR/WebCryptoAPI/#hmac-operations
Chromium: https://chromium.googlesource.com/chromium/src/+/master/components/webcrypto/algorithms/hmac.cc

PR #345, opened 2026-07-27
docs: move PBKDF2 example to a separate file (#283)

bhavit04:
Part of #283, one primitive per PR. Moves the `Pbkdf2SecretKey` example out of the doc comment into `example/webcrypto/pbkdf2/derive_bits.dart` and references it with `{@example}`, same as the HMAC one.

Moving it to a real file surfaced a couple of bugs in the example that I fixed so it actually analyzes and runs:
- `utf8.decode(...)` on the password and salt should be `utf8.encode(...)` (string to bytes, not the other way around)
- `deriveBits` is an instance method, so it's now called on the imported `key` instead of statically on `Pbkdf2SecretKey`

`dart analyze` and `dart format` both pass on the new file.

### DIFF
--- a/example/webcrypto/pbkdf2/derive_bits.dart
+++ b/example/webcrypto/pbkdf2/derive_bits.dart
@@ -1,0 +1,41 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


jonasfj (approved):
Need CLA bot to be happy

PR #348, opened 2026-07-27
fix(ffi): reject overflowing cryptographic parameters

### LINKED ISSUE #347: bug: native FFI wraps oversized PBKDF2 iterations and RSA-PSS salt lengths
## Summary

The native FFI backend silently narrows oversized integer parameters before passing them to BoringSSL. This can change security-sensitive parameters modulo `2^32` instead of rejecting them.

Two affected operations are:

- `Pbkdf2SecretKey.deriveBits`: `iterations = 2^32 + 1` is executed as one iteration.
- RSA-PSS signing and verification: `saltLength = 2^32` is executed as a zero-byte salt.

The browser backend rejects these values during Web IDL normalization because both parameters are defined as `[EnforceRange] unsigned long`.

## Reproduction

```dart

Future<void> main() async {
  final password = await Pbkdf2SecretKey.importRawKey([1, 2, 3, 4]);
  final once = await password.deriveBits(
    256,
    Hash.sha256,
    [5, 6, 7, 8],
    1,
  );
  final wrappedIterations = await password.deriveBits(
    256,
    Hash.sha256,
    [5, 6, 7, 8],
    0x100000001,
  );
  print(_equal(once, wrappedIterations));

  final pair = await RsaPssPrivateKey.generateKey(
    1024,
    BigInt.from(65537),
    Hash.sha256,
  );
  final message = [1, 2, 3, 4];
  final zeroSalt = await pair.privateKey.signBytes(message, 0);
  final wrappedSalt = await pair.privateKey.signBytes(
    message,
    0x100000000,
  );
  print(_equal(zeroSalt, wrappedSalt));
  print(await pair.publicKey.verifyBytes(wrappedSalt, message, 0));
}

bool _equal(List<int> a, List<int> b) {
  if (a.length != b.length) return false;
  for (var i = 0; i < a.length; i++) {
    if (a[i] != b[i]) return false;
  }
  return true;
}
```

Native output:

```text
true
true
true
```

The PBKDF2 call completes immediately and produces exactly the same derived bits as one iteration. The RSA-PSS signature produced with `saltLength = 2^32` is exactly the deterministic zero-salt signature and verifies when `saltLength` is specified as zero.

## Expected behavior

Values outside the Web Crypto `unsigned long` range (`0` through `2^32 - 1`) should be rejected before reaching FFI.

The Web Cryptography specification defines:

- `Pbkdf2Params.iterations` as `required [EnforceRange] unsigned long`.
- `RsaPssParams.saltLength` as `required [EnforceRange] unsigned long`.

https://www.w3.org/TR/WebCryptoAPI/#dfn-Pbkdf2Params
https://www.w3.org/TR/WebCryptoAPI/#dfn-RsaPssParams

## Actual behavior

The native implementations only validate the lower bounds:

- PBKDF2 rejects `iterations <= 0`.
- RSA-PSS rejects `saltLength < 0`.

The values are then passed to 32-bit native parameters, where oversized Dart integers are narrowed modulo `2^32`.

This creates a backend difference: standards-based Web Crypto implementations reject the same inputs with a range error, while the native backend can successfully perform a different and substantially weaker operation.

## Impact

The PBKDF2 case is security-relevant. An application can believe it requested a very expensive password-hardening operation while the native backend performs only a small wrapped iteration count. For example, `2^32 + 1` becomes one iteration.

For RSA-PSS, an oversized salt length can become a smaller salt length, including zero, changing the signature scheme parameters without notifying the caller.

## Suggested fix

- Validate PBKDF2 `iterations` as `1 <= iterations <= 0xffffffff` before the FFI call.
- Validate RSA-PSS `saltLength` as `0 <= saltLength <= 0xffffffff` before signing or verification.
- Add native regression tests at `2^32`, `2^32 + 1`, and valid boundary-adjacent values.
- Audit other Dart integers passed to fixed-width native parameters for the same narrowing behavior.

harrshita123:
Fixes #347

## Summary

- reject PBKDF2 iteration counts that do not fit Web IDL's `unsigned long`
- reject RSA-PSS salt lengths that cannot be represented by BoringSSL's signed 32-bit parameter
- add native regression coverage for integer narrowing at the FFI boundary

## Root cause

The native backend passed arbitrary-width Dart integers directly to fixed-width FFI parameters. Values outside those native ranges were narrowed before reaching BoringSSL, allowing PBKDF2 iterations and RSA-PSS salt lengths to wrap to different values.

The RSA-PSS path also accepted values such as `0xffffffff`, which narrowed to BoringSSL's special `-1` salt-length sentinel.

## Impact

Oversized parameters now fail before the FFI call instead of silently selecting weaker or otherwise different cryptographic parameters. Valid parameter behavior is unchanged.

## Validation

- `dart test test/webcrypto_test.dart` (1,457 tests passed)

### DIFF
--- a/lib/src/impl_ffi/impl_ffi.pbkdf2.dart
+++ b/lib/src/impl_ffi/impl_ffi.pbkdf2.dart
@@ -65,6 +65,7 @@ final class _Pbkdf2SecretKeyImpl implements Pbkdf2SecretKeyImpl {
         'Iterations <= 0 is not allowed for Pbkdf2SecretKey.deriveBits',
       );
     }
+    _checkUnsignedLong(iterations, 'iterations');
     if (length == 0) {
       return Uint8List(0);
     }
--- a/lib/src/impl_ffi/impl_ffi.rsapss.dart
+++ b/lib/src/impl_ffi/impl_ffi.rsapss.dart
@@ -140,6 +140,7 @@ final class _RsaPssPrivateKeyImpl implements RsaPssPrivateKeyImpl {
         'must be a positive integer',
       );
     }
+    _checkInt32(saltLength, 'saltLength');
 
     return _signStream(
       _key,
@@ -219,6 +220,7 @@ final class _RsaPssPublicKeyImpl implements RsaPssPublicKeyImpl {
         'must be a positive integer',
       );
     }
+    _checkInt32(saltLength, 'saltLength');
 
     return _verifyStream(
       _key,
--- a/lib/src/testing/regression/uint32_parameter_overflow.dart
+++ b/lib/src/testing/regression/uint32_parameter_overflow.dart
@@ -1,0 +1,82 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:webcrypto/webcrypto.dart';
+
+const _isFfi = bool.fromEnvironment('dart.library.ffi');
+const _maxUnsignedLong = 0xffffffff;
+
+List<({String name, Future<void> Function() test})> tests() => !_isFfi
+    ? const []
+    : [
+        (
+          name: 'PBKDF2 rejects iteration counts above unsigned long',
+          test: () async {
+            final key = await Pbkdf2SecretKey.importRawKey([1, 2, 3, 4]);
+
+            for (final iterations in [
+              _maxUnsignedLong + 1,
+              _maxUnsignedLong + 2,
+            ]) {
+              await _expectArgumentError(
+                () =>
+                    key.deriveBits(256, Hash.sha256, [5, 6, 7, 8], iterations),
+                'Expected PBKDF2 iterations $iterations to be rejected',
+              );
+            }
+          },
+        ),
+        (
+          name: 'RSA-PSS rejects salt lengths that narrow at the FFI boundary',
+          test: () async {
+            final pair = await RsaPssPrivateKey.generateKey(
+              1024,
+              BigInt.from(65537),
+              Hash.sha256,
+            );
+            final message = [1, 2, 3, 4];
+            final signature = await pair.privateKey.signBytes(message, 0);
+
+            for (final saltLength in [
+              0x80000000,
+              _maxUnsignedLong,
+              _maxUnsignedLong + 1,
+              _maxUnsignedLong + 2,
+            ]) {
+              await _expectArgumentError(
+                () => pair.privateKey.signBytes(message, saltLength),
+                'Expected RSA-PSS signing salt length $saltLength to be rejected',
+              );
+
+              await _expectArgumentError(
+                () =>
+                    pair.publicKey.verifyBytes(signature, message, saltLength),
+                'Expected RSA-PSS verification salt length $saltLength to be rejected',
+              );
+            }
+          },
+        ),
+      ];
+
+Future<void> _expectArgumentError(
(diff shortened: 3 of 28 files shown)


lib/src/webcrypto/webcrypto.pbkdf2.dart
```diff
@@ -82,6 +82,7 @@ final class Pbkdf2SecretKey {
   /// an exhaustive search for the derived key, but it will also make the
   /// key derivation operation slower. For details on [iterations] see
   /// [RFC 8018 section 4.2][1].
+  /// The [iterations] count must be between `1` and `2^32 - 1`.
```
jonasfj: Don't document this, it's absurd. It'll never finish.

lib/src/webcrypto/webcrypto.rsapss.dart
```diff
@@ -248,6 +248,7 @@ final class RsaPssPrivateKey {
   /// The [saltLength] is typically zero or length of the [Hash], for a
   /// discussion of appropriate values for [saltLength] see
   /// [RFC 3447 Section 9.1, Notes 4][1].
+  /// The [saltLength] must be between `0` and `2^32 - 1`.
```
jonasfj: Don't document! it's a sanity limit!

lib/src/testing/regression/uint32_parameter_overflow.dart
```diff
@@ -0,0 +1,82 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:webcrypto/webcrypto.dart';
+
+const _isFfi = bool.fromEnvironment('dart.library.ffi');
+const _maxUnsignedLong = 0xffffffff;
+
+List<({String name, Future<void> Function() test})> tests() => !_isFfi
```
jonasfj, line 22: why? because the tests don't work in JS?

lib/src/impl_ffi/impl_ffi.pbkdf2.dart
```diff
@@ -65,6 +65,7 @@ final class _Pbkdf2SecretKeyImpl implements Pbkdf2SecretKeyImpl {
         'Iterations <= 0 is not allowed for Pbkdf2SecretKey.deriveBits',
       );
     }
+    _checkUnsignedLong(iterations, 'iterations');
```
jonasfj, line 68: This is fine as a sanity limit, nobody will ever hit this, if you do the half of the limit it'll probably never finish

lib/src/impl_ffi/impl_ffi.rsapss.dart
```diff
@@ -140,6 +140,7 @@ final class _RsaPssPrivateKeyImpl implements RsaPssPrivateKeyImpl {
         'must be a positive integer',
       );
     }
+    _checkInt32(saltLength, 'saltLength');
```
jonasfj, line 143: Sure fine sanity check.

lib/src/webcrypto/webcrypto.pbkdf2.dart
harrshita123: I removed the PBKDF2 upper-bound sentence. The check remains an internal sanity guard at the FFI boundary.

lib/src/webcrypto/webcrypto.rsapss.dart
harrshita123: I removed the RSA-PSS limit from both signing method docs; it remains internal sanity validation.

lib/src/testing/regression/uint32_parameter_overflow.dart
harrshita123: Yes. Browser backends already enforce these ranges during Web IDL conversion, while these cases specifically exercise integer narrowing in native FFI. I added a short comment beside the guard to make that distinction clear.

lib/src/impl_ffi/impl_ffi.pbkdf2.dart
harrshita123: I’ve kept this as an internal sanity check and removed the public limit documentation.

lib/src/impl_ffi/impl_ffi.rsapss.dart
harrshita123: I left the signed 32-bit sanity check in place and removed the public range documentation.

PR #350, opened 2026-07-29
fix(hkdf): validate output length before allocation

### LINKED ISSUE #349: bug: native HKDF allocates output before validating the RFC 5869 length limit
## Summary

The native FFI implementation of `HkdfSecretKey.deriveBits` allocates the requested output buffer before validating HKDF's maximum output length. An oversized request can therefore trigger an arbitrarily large native allocation attempt before it is rejected.

RFC 5869 limits HKDF output to `255 * HashLen` bytes. The implementation already translates BoringSSL's `HKDF_R_OUTPUT_TOO_LARGE` error into a useful `OperationError`, but BoringSSL is invoked only after the output buffer has been allocated.

## Reproduction

```dart

Future<void> main() async {
  final key = await HkdfSecretKey.importRawKey([1, 2, 3, 4]);

  for (final length in <int>[
    (255 * 32 + 1) * 8,
    9223372036854775800,
  ]) {
    try {
      await key.deriveBits(length, Hash.sha256, const [], const []);
    } catch (error) {
      print('$length: $error');
    }
  }
}
```

Native output:

```text
65288: Length specified for HkdfSecretKey.deriveBits is too long
9223372036854775800: error:0e000041:common libcrypto routines:OPENSSL_internal:malloc failure
```

The first value is one byte beyond SHA-256 HKDF's maximum and reaches BoringSSL's intended output-length check. The second value fails earlier while attempting to allocate the output buffer.

## Expected behavior

Every output length greater than `255 * HashLen` bytes should be rejected before allocating output memory, with the existing deterministic "too long" `OperationError`.

RFC 5869 Section 2.3 defines `L <= 255 * HashLen`:

https://www.rfc-editor.org/rfc/rfc5869#section-2.3

The Web Cryptography HKDF derive-bits algorithm delegates to that HKDF extract-and-expand operation:

https://www.w3.org/TR/WebCryptoAPI/#hkdf-operations

## Actual behavior

The FFI implementation currently performs these operations in this order:

```dart
final lengthInBytes = length ~/ 8;

return _Scope.async((scope) async {
  final out = scope<ffi.Uint8>(lengthInBytes);
  final r = ssl.HKDF(
    out,
    lengthInBytes,
  );
```

Consequently, the requested native allocation occurs before `ssl.HKDF` can report `HKDF_R_OUTPUT_TOO_LARGE`.

## Impact

Invalid input that should be rejected cheaply can instead cause very large native allocation attempts. Depending on allocator and process memory conditions, this can produce inconsistent errors, memory pressure, or process termination.

This is especially relevant when the requested derivation length is influenced by external configuration or input.

## Suggested fix

- Determine the selected hash's output size before entering the allocation scope.
- Reject `lengthInBytes > 255 * hashLengthInBytes` with the existing "too long" `OperationError`.
- Keep the BoringSSL `HKDF_R_OUTPUT_TOO_LARGE` handling as a defensive fallback.
- Add regression tests covering:
  - the exact maximum for SHA-1, SHA-256, SHA-384, and SHA-512;
  - one byte beyond each maximum; and
  - a very large valid positive Dart integer, confirming rejection occurs without an allocation failure.

harrshita123:
## Summary

- validate the RFC 5869 HKDF output limit before allocating the native output buffer
- preserve BoringSSL's output-too-large handling as a defensive fallback
- cover the exact limit and the first invalid length for every supported hash
- verify that a very large length is rejected without reaching native allocation

## Root cause

The FFI implementation allocated `length ~/ 8` bytes before calling BoringSSL. BoringSSL enforces HKDF's `255 * HashLen` maximum, but an extremely large request could fail during allocation before reaching that validation.

The implementation now derives the maximum from `EVP_MD_size` and rejects oversized output lengths before entering the allocation scope.

## Validation

- `dart format --output none --set-exit-if-changed lib/src/impl_ffi/impl_ffi.hkdf.dart test/hkdf_output_length_test.dart`
- `dart test test/hkdf_output_length_test.dart`
- `dart test -p vm --exclude-tags browser-interop`

Fixes #349

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # 0.6.2-wip
 * Fixed JS interop to enable WebAssembly.
 * Fixed native RSA public-key JWK imports to reject private key material.
+* Fixed native HKDF output-length validation to reject oversized requests
+  before allocating memory.
 
 # 0.6.1
 * Added Dart native build hooks and native asset lookup for the bundled
--- a/lib/src/impl_ffi/impl_ffi.hkdf.dart
+++ b/lib/src/impl_ffi/impl_ffi.hkdf.dart
@@ -57,6 +57,12 @@ final class _HkdfSecretKeyImpl implements HkdfSecretKeyImpl {
     }
 
     final lengthInBytes = length ~/ 8;
+    const lengthTooLong =
+        'Length specified for HkdfSecretKey.deriveBits is too long';
+    final maxLengthInBytes = 255 * ssl.EVP_MD_size(md);
+    if (lengthInBytes > maxLengthInBytes) {
+      throw operationError(lengthTooLong);
+    }
 
     return _Scope.async((scope) async {
       final out = scope<ffi.Uint8>(lengthInBytes);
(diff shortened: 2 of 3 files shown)


HamdaanAliQuatil (approved):
thanks :)

PR #351, opened 2026-07-30
feat(impl_jni): add JCA ECDSA implementation

mfazrinizar:
## Summary

Adds the JNI/JCA ECDSA implementation on top of the Android JCA backend.

This PR implements:
- P-256, P-384, and P-521 key generation
- PKCS#8 private-key and SPKI/raw public-key import/export
- ECDSA JWK import/export with key-pair validation
- byte and stream signing and verification for every backend hash
- strict conversion between JCA DER signatures and Web Crypto raw signatures
- persistent isolate-unsendable JCA key ownership
- focused desktop JNI interoperability tests and an Android smoke test

## Known limitations

- A valid EC private PKCS#8 key without an embedded public point is
  provider-validated but currently rejected with `UnsupportedError`. Standard
  JCA does not expose the public coordinates needed for private JWK export.

## Testing

```sh
dart run jni:setup
```

Verified with:

```sh
dart analyze
dart test test/impl_jni_ecdsa_test.dart
dart test test/webcrypto_test.dart -p vm -n ECDSA
flutter test integration_test/jni_ecdsa_test.dart -d emulator-name
cd ../..
git diff --check
```

### DIFF
--- a/example/webcrypto_demo_flutter_app/integration_test/jni_ecdsa_test.dart
+++ b/example/webcrypto_demo_flutter_app/integration_test/jni_ecdsa_test.dart
@@ -41,4 +41,17 @@ void main() {
       );
     }
   });
+
+  testWidgets('public API imports point-less P-256 PKCS#8', (_) async {
+    final privateKey = await EcdsaPrivateKey.importPkcs8Key(
+      base64.decode(
+        'MEECAQAwEwYHKoZIzj0CAQYIKoZIzj0DAQcEJzAlAgEBBCCdKfq9jvibt61Gf8H5'
+        'nePEVusJphstWiK4ELnhbBr8bQ==',
+      ),
+      EllipticCurve.p256,
+    );
+    final jwk = await privateKey.exportJsonWebKey();
+    expect(jwk['x'], 'IhPWW8AGjzkxOj2gMtMDBO9nlVrzvD7Qm6AtAl6X8sE');
+    expect(jwk['y'], 'n2lKzqEjootvel5634HoRtxO0GlwA7UBDa477Cd4Qto');
+  });
 }
--- a/example/webcrypto_demo_flutter_app/integration_test/jni_ecdsa_test.dart
+++ b/example/webcrypto_demo_flutter_app/integration_test/jni_ecdsa_test.dart
@@ -1,0 +1,44 @@
+import 'dart:convert';
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+import 'package:webcrypto/webcrypto.dart';
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  testWidgets('public API ECDSA supports every named curve', (_) async {
+    for (final testCase
+        in <({EllipticCurve curve, Hash hash, int signatureLength})>[
+          (curve: EllipticCurve.p256, hash: Hash.sha256, signatureLength: 64),
+          (curve: EllipticCurve.p384, hash: Hash.sha384, signatureLength: 96),
+          (curve: EllipticCurve.p521, hash: Hash.sha512, signatureLength: 132),
+        ]) {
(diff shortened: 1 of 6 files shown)


lib/src/impl_jni/impl_jni.ec_common.dart
```diff
@@ -0,0 +1,907 @@
...
+      final encoded = arena.copyToJByteArray(keyData);
+      final keySpec = PKCS8EncodedKeySpec(encoded)..releasedBy(arena);
+      final key = keyFactory.generatePrivate(keySpec);
+      if (key == null) {
+        throw AssertionError('JCA EC KeyFactory returned a null private key');
+      }
+
+      return _validateEcKeyBeforeOwnershipTransfer(key, () {
+        _validateEcPrivateKey(arena, key, curve);
+        // TODO: Decide whether point-less EC private PKCS#8 keys need a
```
HamdaanAliQuatil: blocking: this rejects valid EC PKCS#8 input. RFC 5915 makes the inner `ECPrivateKey.publicKey [1]` optional; it can be omitted when the corresponding public key is distributed separately. the existing FFI backend deliberately derives the public key when it is omitted.

this backend accepts and validates the private scalar, then throws only because it cannot retain/export `x` and `y`. that changes the public `importPkcs8Key` behavior by backend. 

please add a supported public-point derivation path, or bring an explicit compatibility decision to us before narrowing the accepted PKCS#8 inputs.

please also add a regression case using a valid p-256 PKCS#8 key without `[1] publicKey`.

see https://www.rfc-editor.org/rfc/rfc5915#section-3

HamdaanAliQuatil (changes requested):

lib/src/impl_jni/impl_jni.ec_common.dart
mfazrinizar: After rebasing onto #353, RFC 5915 EC private keys without the optional public point are now supported.

The backend computes Q = d * G through JCA KeyAgreement using the curve generator. Dart only reconstructs the two public y candidates from the resulting x, and JCA ECDSA verification selects the matching point, so private-scalar multiplication remains inside JCA.

I also added fixed PKCS#8 regressions for P-256, P-384, and P-521, plus Android P-256 coverage.

example/webcrypto_demo_flutter_app/integration_test/jni_ecdsa_test.dart
```diff
@@ -0,0 +1,57 @@
...
+
+      expect(signature, hasLength(testCase.signatureLength));
+      expect(
+        await publicKey.verifyBytes(signature, data, testCase.hash),
+        isTrue,
+      );
+    }
+  });
+
+  testWidgets('public API imports point-less P-256 PKCS#8', (_) async {
```
HamdaanAliQuatil: this path is now provider-specific: it interprets ECDH `generateSecret()` as the affine x coordinate, then recovers y with curve-specific arithmetic. the desktop test covers all three curves, but Android currently exercises this path only for p-256.

please run the same point-less PKCS#8 fixtures for p-384 and p-521 in this integration test too, asserting the derived JWK coordinates and sign/verify. p-521 in particular is important because of its 66-byte coordinate encoding.

HamdaanAliQuatil (changes requested):

example/webcrypto_demo_flutter_app/integration_test/jni_ecdsa_test.dart
mfazrinizar: Added the same P-256, P-384, and P-521 point-less PKCS#8 fixtures to the Android integration test. It now asserts the exact derived JWK coordinates, verifies the 32/48/66-byte coordinate widths explicitly, and runs sign/verify with the corresponding curve hash. Tested on Android 16 x86_64 emulator, both ECDSA integration tests pass.

HamdaanAliQuatil (approved):

PR #352, opened 2026-07-31
test(ecdh): enable P-521 tests on Safari

### LINKED ISSUE #129: Tests if P-521 works on Safari
Comments in our code reference a bug claiming that P-521 doesn't work in safari:
https://github.com/google/webcrypto.dart/blob/e7fbe5012c83fb1907901f4ccb03d816bd574f7d/lib/src/webcrypto/webcrypto.dart#L71-L76

However, reading: https://bugs.webkit.org/show_bug.cgi?id=216755 this appears to have been fixed.

We should:
 * [ ] Check if this is true,
 * [ ] Enable tests covering P-521 on Safari / Mac
 * [ ] Add additional tests if warranted.
 * [ ] Remove the comment.

harrshita123:
## Summary

- enable the existing P-521 ECDH vectors when the test suite runs on Safari
- remove the obsolete documentation claiming that Safari does not support P-521
- remove the runtime-detection import that is no longer needed by the ECDH tests

Safari now supports P-521, so excluding these vectors leaves a supported path untested. The existing vectors already cover key import and export across raw, SPKI, PKCS8, and JWK formats, as well as 528-bit key derivation.

Closes #129.

## Validation

- full VM test suite (1,448 passed; 1 platform-specific skip)
- P-521 ECDH tests on Safari 26.5.2 with Dart2JS (41 passed)
- P-521 ECDH tests on Chrome with Dart2JS and Dart2Wasm (82 passed)

### DIFF
--- a/lib/src/impl_interface/impl_interface.dart
+++ b/lib/src/impl_interface/impl_interface.dart
@@ -40,17 +40,7 @@ typedef KeyPair<T, S> = ({T privateKey, S publicKey});
 ///
 /// > [!NOTE]
 /// > Additional values may be added to this enum in the future.
-enum EllipticCurve {
-  p256,
-  p384,
-
-  ///
-  ///
-  /// P-521 is **not supported on Safari**, see [bug 216755 (bugs.webkit.org)][1].
-  ///
-  /// [1]: https://bugs.webkit.org/show_bug.cgi?id=216755
-  p521,
-}
+enum EllipticCurve { p256, p384, p521 }
 
 /// Thrown when an operation failed for an operation-specific reason.
 final class OperationError extends Error {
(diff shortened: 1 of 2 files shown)


jonasfj (approved):

jonasfj:
Awesome, it might be worth exploring if we can get Github Actions to cover Safari again, that would be a separate PR.
And probably one we should retrigger a few times to make sure Safari isn't flaky, I think it had a tendency to hang and that's why we eventually decided to disable testing in Safari.

harrshita123:
> Awesome, it might be worth exploring if we can get Github Actions to cover Safari again, that would be a separate PR.
> And probably one we should retrigger a few times to make sure Safari isn't flaky, I think it had a tendency to hang and that's why we eventually decided to disable testing in Safari.

Thanks! I’ll explore re-enabling Safari coverage in GitHub Actions separately and run the workflow several times to check for hangs or flaky behavior before opening a PR.

PR #359, opened 2026-08-07
Validate HMAC key generation length before allocation

### LINKED ISSUE #358: Native HMAC key generation allocates memory before validating the Web Crypto length range
The native implementation of `HmacSecretKey.generateKey` accepts arbitrarily large positive `length` values.

The public API currently rejects only zero and negative values. The native backend then calculates the required byte count and constructs a `Uint8List` immediately:

```dart
final keyData = Uint8List((length / 8).ceil());
```

Consequently, an out-of-range value can cause an enormous allocation attempt instead of being rejected deterministically.

### Reproduction

```dart
await HmacSecretKey.generateKey(
  Hash.sha256,
  length: 9007199254740992,
);
```

### Expected behavior

The value should be rejected before allocation because [`HmacKeyGenParams.length`](https://www.w3.org/TR/webcrypto/#dfn-HmacKeyGenParams) is defined as an `[EnforceRange] unsigned long`.

### Actual behavior

The native implementation attempts to allocate storage based on the unvalidated value, potentially resulting in an out-of-memory failure.

### Suggested fix

- Validate explicit HMAC key lengths against the Web Crypto/WebIDL range before dispatching to a backend.
- Preserve the existing rejection of zero and negative lengths.
- Add regression tests for the upper boundary and an extremely large positive value.
- Avoid allocating a maximum-size key in the boundary tests.

harrshita123:
## Summary

- reject explicit HMAC key lengths outside the WebIDL `unsigned long` range before backend dispatch
- calculate the native key byte length using integer ceiling division
- add regression coverage for the first out-of-range value and a very large value

## Motivation

`HmacSecretKey.generateKey` previously validated only zero and negative lengths. On the native backend, an arbitrarily large positive value reached `Uint8List` allocation, potentially causing an out-of-memory failure instead of a deterministic argument error.

Validating the upper bound in the shared API keeps backend behavior consistent with the `[EnforceRange] unsigned long` type used by Web Crypto.

## Testing

- `dart analyze`
- `dart test -p vm`
- `dart test -p chrome -c dart2js`
- focused regression test with Chrome/dart2wasm

Fixes #358

### DIFF
--- a/lib/src/impl_ffi/impl_ffi.hmac.dart
+++ b/lib/src/impl_ffi/impl_ffi.hmac.dart
@@ -75,7 +75,7 @@ Future<HmacSecretKeyImpl> hmacSecretKey_generateKey(
 }) async {
   final h = _HashImpl.fromHash(hash);
   length ??= ssl.EVP_MD_size(h._md) * 8;
-  final keyData = Uint8List((length / 8).ceil());
+  final keyData = Uint8List((length + 7) ~/ 8);
   fillRandomBytes(keyData);
 
   return _HmacSecretKeyImpl(_asUint8ListZeroedToBitLength(keyData, length), h);
--- a/lib/src/impl_ffi/impl_ffi.hmac.dart
+++ b/lib/src/impl_ffi/impl_ffi.hmac.dart
@@ -75,7 +75,7 @@ Future<HmacSecretKeyImpl> hmacSecretKey_generateKey(
 }) async {
   final h = _HashImpl.fromHash(hash);
   length ??= ssl.EVP_MD_size(h._md) * 8;
-  final keyData = Uint8List((length + 7) ~/ 8);
+  final keyData = Uint8List((length / 8).ceil());
   fillRandomBytes(keyData);
 
   return _HmacSecretKeyImpl(_asUint8ListZeroedToBitLength(keyData, length), h);
--- a/lib/src/webcrypto/webcrypto.hmac.dart
+++ b/lib/src/webcrypto/webcrypto.hmac.dart
@@ -197,6 +197,9 @@ final class HmacSecretKey {
     if (length != null && length <= 0) {
       throw ArgumentError.value(length, 'length', 'must be positive');
     }
+    if (length != null && length > 0xffffffff) {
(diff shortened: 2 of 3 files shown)


lib/src/webcrypto/webcrypto.hmac.dart
```diff
@@ -197,6 +197,9 @@ final class HmacSecretKey {
     if (length != null && length <= 0) {
       throw ArgumentError.value(length, 'length', 'must be positive');
     }
+    if (length != null && length > 0xffffffff) {
+      throw ArgumentError.value(length, 'length', 'must be at most 4294967295');
```
jonasfj: This is really just an extreme sanity check.
```suggestion
      throw ArgumentError.value(length, 'length', 'too large');
```

Let's not promise a specific number.

lib/src/impl_ffi/impl_ffi.hmac.dart
```diff
@@ -75,7 +75,7 @@ Future<HmacSecretKeyImpl> hmacSecretKey_generateKey(
 }) async {
   final h = _HashImpl.fromHash(hash);
   length ??= ssl.EVP_MD_size(h._md) * 8;
-  final keyData = Uint8List((length / 8).ceil());
+  final keyData = Uint8List((length + 7) ~/ 8);
```
jonasfj: Does this change behavior?

jonasfj: I don't mind as such, but why?

lib/src/webcrypto/webcrypto.hmac.dart
harrshita123: I've changed the message to `too large` so the error doesn't promise a specific limit.

lib/src/impl_ffi/impl_ffi.hmac.dart
harrshita123: The calculation was equivalent for accepted lengths, but it wasn't necessary for this fix. I've reverted it to keep the change focused.

HamdaanAliQuatil (approved):

Issue #371, opened 2026-08-21
[Android JCA] Support RSA PKCS#8 imports when providers hide CRT parameters


mfazrinizar:
## Problem

On the tested Android API 24-26 providers, `KeyFactory` accepts a valid RSA PKCS#8 encoding but returns a private-key object that does not implement `RSAPrivateCrtKey`. The JNI backend then rejects the key because its import-time validation and private JWK export require the CRT components.

The tested API 36 provider returns an `RSAPrivateCrtKey` for the same generate, encode, and re-import path.

## Provider Evidence

On the Android API 24 x86_64 image with AndroidOpenSSL 1.0:

- RSA generation returns `OpenSSLRSAPrivateCrtKey`.
- Re-importing the generated key's PKCS#8 encoding returns `OpenSSLRSAPrivateKey`.
- The imported key does not implement `RSAPrivateCrtKey`.
- `KeyFactory.getKeySpec(..., RSAPrivateCrtKeySpec.class)` returns a plain `RSAPrivateKeySpec`, so the provider still does not expose the CRT values.

API 25 and 26 reproduce the same `package:webcrypto` PKCS#8 import failure. API 36 re-imports the generated encoding as `OpenSSLRSAPrivateCrtKey` and passes the shared RSA-PSS suite with the parameter-order fix.

The shared RSA key layer affects RSASSA-PKCS1-v1_5, RSA-OAEP, and RSA-PSS.

## Constraint

The backend checks CRT consistency during import so malformed private keys fail as `FormatException` before use. Private JWK export requires `n`, `e`, `d`, `p`, `q`, `dp`, `dq`, and `qi`. Removing the CRT requirement would weaken import validation and leave private JWK export incomplete.

## Decision Needed

Choose one source for provider-independent RSA components:

- Add a small reviewed PKCS#8/DER parser to the JNI backend and retain the parsed components beside the persistent JCA key.
- Use a reviewed ASN.1 dependency for the same parse-and-retain design.
- Restrict the supported Android provider/API matrix and document the limitation.

The API 24 experiment rules out direct CRT extraction through `KeyFactory.getKeySpec`.

## Acceptance Criteria

- Valid generated and supplied PKCS#8 keys import on the selected Android provider/API matrix.
- Malformed RSA keys fail during import with `FormatException`.
- PKCS#8 and private JWK exports preserve the imported key material.
- Persistent JCA key ownership and isolate-unsendable behavior remain intact.
[...]

PR #375, opened 2026-08-23
fix(impl_jni): configure RSA-PSS parameters after initialization

mfazrinizar:
## Summary

- Configure `PSSParameterSpec` after `Signature.initSign` or `Signature.initVerify`.
- Preserve the requested digest, MGF1 digest, salt length, and trailer field.
- Fix verification of externally generated signatures that use non-default salt lengths on the tested Android provider.

## Root Cause

The tested Android JCA provider resets its active RSA-PSS parameters during signature initialization. The previous order configured `PSSParameterSpec` before initialization, so the provider replaced the requested values with its defaults.

The backend now obtains and initializes `Signature` before applying the requested parameters. This PR does not change generated bindings or persistent key ownership.

Issue #371 tracks API 24-26 providers that return non-CRT private-key wrappers and block the shared RSA tests before the PSS operation runs.

## Testing

Desktop JNI setup, if needed:

- `dart run jni:setup`

Verified with:

- `dart format --output=none --set-exit-if-changed lib/src/impl_jni/impl_jni.rsapss.dart`
- `git diff --check`
- `dart analyze lib/src/impl_jni/impl_jni.rsapss.dart test/impl_jni_rsapss_test.dart`
- `dart test test/impl_jni_rsapss_test.dart`
- `dart test test/webcrypto_test.dart -p vm -n 'RSA-PSS'`
- `flutter test integration_test/webcrypto_test.dart -d emulator-name --name 'RSA-PSS'` from `example/webcrypto_demo_flutter_app`

The focused desktop suite passes 3 cases, the shared desktop RSA-PSS selection passes 366 cases, and the tested Android API 36 x86_64 provider passes all 366 shared RSA-PSS cases. API 24-26 remain blocked by the separate RSA private-key import issue.

Follow-up to #341.

### DIFF
--- a/lib/src/impl_jni/impl_jni.rsapss.dart
+++ b/lib/src/impl_jni/impl_jni.rsapss.dart
@@ -169,8 +169,11 @@ Future<Uint8List> _signRsaPssStream(
 ) async {
   final arena = jni.Arena();
   try {
-    final signature = _createRsaPssSignature(arena, hash, saltLength);
+    final signature = _getRsaPssSignature(arena, hash);
     signature.initSign(key.key);
+    // The tested Android JCA provider resets PSS parameters during
+    // initialization.
+    _configureRsaPssSignature(arena, signature, hash, saltLength);
 
     final buffer = jni.JByteArray(_defaultChunkSize)..releasedBy(arena);
     await for (final chunk in data) {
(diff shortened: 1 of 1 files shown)


HamdaanAliQuatil (approved):
lgtm :)

PR #398, opened 2026-09-09
docs: extract RSA-OAEP PKCS#8 import example

Tiancheng-Xu:
## Summary

- Move the existing `RsaOaepPrivateKey.importPkcs8Key` dartdoc example to `example/webcrypto/rsa_oaep/import_pkcs8_key.dart`.
- Replace the inline block with the corresponding `{@example ...#example}` directive.
- Preserve the example behavior and keep the change limited to this method.

Part of #283.

## Validation

- `git diff --check`
- Structural extraction check: example body preserved, reference appears once, old inline block removed, region markers balanced, and Apache license header present.
- `dart format` and `dart analyze` were not run locally because Dart/Flutter are not installed in the environment; upstream CI will provide the repository toolchain validation.


example/webcrypto/rsa_oaep/import_pkcs8_key.dart
```diff
@@ -0,0 +1,41 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// ignore_for_file: avoid_print
+
+// #region example
+import 'package:webcrypto/webcrypto.dart';
+import 'package:pem/pem.dart';
+
```
HamdaanAliQuatil, line 20: this file currently fails dart analyze because pem isn’t declared in pubspec.yaml. the missing PemCodec and PemLabel errors follow from that. could you add pem as a dev dependency so the extracted example is actually analyzable?

HamdaanAliQuatil (changes requested):

Tiancheng-Xu:
Follow-up: CI now passes `check-changes` and `cla/google` on commit `49a4fd2`. The extracted example imports `package:pem/pem.dart`, so I added `pem: ^2.0.6` under `dev_dependencies`; this keeps the package runtime dependency surface unchanged while allowing `dart analyze` to resolve the example. Could you please re-review the updated commit?

Tiancheng-Xu:
Follow-up verified: `pem: ^2.0.6` is now present under `dev_dependencies`, so the extracted example resolves `PemCodec`/`PemLabel` without expanding runtime dependencies. Current visible checks (`check-changes`, `cla/google`) pass; the earlier Format + Lints failure was from the prior run and is no longer present in the current check set.

HamdaanAliQuatil (approved):

